PR #8, opened 2022-05-11
Remove devShell. Use devShells.default instead.

roberth:


jmatsushita:
Hi @roberth 

I'm giving `flake-modules-core` a try to see if it can restore my sanity regarding dealing with `system` :) 

I bumped into the issue however that `nix develop` without argument defaults to `devShell.<system>` and doesn't know about `devShells.default` AFAICT. So `nix develop .#default` works, but not `nix develop` with this change.

Maybe that's an unintended consequence? 

Cheers!

Jun

jmatsushita:
Nevermind 😓 I was using nix `2.4` and the current stable `2.8.1` supports this 👍

PR #9, opened 2022-05-13
Docs and initial ci

roberth:


Mic92:
I think this broke flake. I.e. when running the template `template/default`:

╭─[/tmp/tmp.41XtcGzWLa/flake-modules-core/template/default]─[main*]─[bisect]
╰─ % nix build .#hello
warning: Git tree '/tmp/tmp.41XtcGzWLa/flake-modules-core' is dirty
error: value is a function while a set was expected

       at /nix/store/57zwl7b883kkya60akvg6wssl1qkxwl5-source/lib/modules.nix:689:24:

          688|           t' = opt.options.type;
          689|           mergedType = t.typeMerge t'.functor;
             |                        ^
          690|           typesMergeable = mergedType != null;
(use '--show-trace' to show detailed location information)

roberth:
It needs https://github.com/NixOS/nixpkgs/pull/172813.
Let me check if I can polyfill it.

roberth:
Done in #14. May need to do `functionTo` calls in less convenient locations later, in case e.g. `overlays` becomes a problem.

PR #15, opened 2022-05-17
Deploy basic site with netlify

roberth:
Giving https://github.com/hercules-ci/hercules-ci-effects/pull/62 a spin


Kranzes:
Is there any reason for adding netlify-cli to nativeBuildInputs? It is uneeded with effects.netlifyDeploy.

PR #17, opened 2022-05-18
introduce lib.mkFlake

zimbatm:
This makes the most common use-case a tiny bit less verbose.

I didn't name it mkFlakeModule because it would be redundant with the project name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,18 +34,15 @@ Otherwise, add the input,
     flake-modules-core.inputs.nixpkgs.follows = "nixpkgs";
 ```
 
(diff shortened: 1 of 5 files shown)


roberth (commented):
What a nice little improvement!

This doesn't seem to lose power compared to the wrapped function, as unexposed `evalModules` results can be passed through via new / temporary / custom options.

It seems that the other remaining uses can also be replaced similarly.

Note to self: add bors bot, add tests for the templates.

zimbatm:
Let me know if there is more to do

roberth:
Thanks!

PR #28, opened 2022-05-28
perSystem: add a default value for `systems`

### LINKED ISSUE #25: Add `mkSimpleFlake`?
In https://github.com/srid/haskell-template I'm using `flake-parts` in place of `flake-utils`. The minimal code I came up with in order to replace the latter's `eachDefaultSystem` is as follows:

```nix
  outputs = { self, nixpkgs, flake-parts, ... }:
    flake-parts.lib.mkFlake { inherit self; } {
      systems = nixpkgs.lib.systems.flakeExposed;
      perSystem = { self', inputs', pkgs, system, ... }:
        ..
```

https://github.com/srid/haskell-template/blob/3fc6858830ecee3d2fe1dfe9a8bfa2047cf561ac/flake.nix#L14-L19

Since I imagine most non-monorepos using `flake-parts` to repeat the above verbatim, would it be worthwhile to expose a function that obviates both `systems` and `perSystem`? i.e.

```nix
  outputs = { self, flake-parts, ... }:
    flake-parts.lib.mkSimpleFlake { inherit self; } ({ self', inputs', pkgs, system, ... }: 
      ...
    );

ie., `mkSimpleFlake` is like `mkFlake` but uses all systems built by Hydra, as well as assumes the user specifies a single-module flake.

### PR DESCRIPTION
**Pull request by srid**
Resolves #25 - specifically https://github.com/hercules-ci/flake-parts/issues/25#issuecomment-1140038700

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,6 @@
+# Now
+
+- `perSystem`: Set `systems` argument by default to `nixpkgs.lib.systems.flakeExposed`
 
 # 2022-05-25
 
--- a/modules/perSystem.nix
+++ b/modules/perSystem.nix
@@ -18,6 +18,7 @@ in
     systems = mkOption {
       description = "All the system types to enumerate in the flake.";
       type = types.listOf types.str;
+      default = lib.systems.flakeExposed;
     };
 
     perInput = mkOption {

### CODE REVIEW

**Review by roberth — COMMENTED** [review]
After some discussion and researching the `flakeExposed` attribute, I don't think we should have this default here. Hang tight for a solution though.

The `flakeExposed` attribute is a very recent addition to Nixpkgs, which makes me a bit uneasy about the stability of this (while acknowledging that flakes are entirely unstable).

Setting this default also goes against how Eelco intended `systems` to be used: to provide known-working attributes. By setting this default, we remove the "known" aspect and more likely than not, the "working" aspect. That would be bad for your users and all `flake-parts` users.

`flake-parts` is intended to be a common core for many distributed expressions to interoperate. Bringing an opinionated default into the core defeats its purpose as a unifying interface, as it gives others reasons to reject it.

As an example, I've only accepted the default `pkgs` because I expect almost everyone to agree on its definition and the existence of a `pkgs` module argument being a unifying force, until a better solution is applied (`nixpkgs.flakeModules.pkgs`, TBD).

But, while I'm sorry for the bad news part, there is a solution, which is to create your own module of opinionated defaults. If this single line is really important, presumably you have it in more than one repo, so it might make sense to create a module with personal defaults in a separate repo. You can use that not just to set the default `systems`, but also a default Nix formatter with [`pre-commit-hooks-nix`](https://github.com/cachix/pre-commit-hooks.nix/pull/130) for example.

**Comment by srid** [comment]
@roberth No worries.

> If this single line is really important, presumably you have it in more than one repo, so it might make sense to create a module with personal defaults in a separate repo. You can use that not just to set the default `systems`, but also a default Nix formatter with [`pre-commit-hooks-nix`](https://github.com/cachix/pre-commit-hooks.nix/pull/130) for example.

Is there documentation on 'creating modules for flake-parts'? The linked PR has `options` and `config`. What are these? Are there other fields? Its template flake sets `pre-commit.settings` but when using it, instead it accesses `config.pre-commit..`; why is that? I suppose my (beginner) questions are also relevant for https://github.com/srid/haskell-template/issues/21

**Comment by roberth** [comment]
> Is there documentation on 'creating modules for flake-parts'?

No docs for writing flake-parts modules specifically yet, but the general structure like `config`, `options`, `mkOption` and the `types` is the same as that of NixOS modules, which are documented in the [NixOS manual](https://nixos.org/manual/nixos/stable/index.html#sec-writing-modules).

Instead of NixOS options like `services.<name>.*`, `boot.*` and `systemd.*`, we have [flake options](https://flake.parts/options.html).

> Its template flake sets `pre-commit.settings` but when using it, instead it accesses `config.pre-commit..`; why is that?

When you write a simple module, it's often possible to omit `config.`. These modules are equivalent

```nix
{
  systems = ["x86_64-linux"];
}
```

```nix
{
  config.systems = ["x86_64-linux"];
}
```

This is also part of the reason why you don't have to use `config` in submodules.

PR #33, opened 2022-06-03
Set packages.default for `nix run`

edrex:
With the pre-2.7 `defaultPackage` key, one would write
```
defaultPackage = packages.hello
```

I guess you can do the same in 2.7, but then you have two attrs in packages, which seems like it might confuse tooling:

```
packages.default = packages.hello
```

I wonder if packages.default shouldn't be a string? but then it would need validation...

### DIFF
--- a/template/default/flake.nix
+++ b/template/default/flake.nix
@@ -23,7 +23,7 @@
         # system.
 
         # Equivalent to  inputs'.nixpkgs.legacyPackages.hello;
-        packages.hello = pkgs.hello;
+        packages.default = pkgs.hello;
       };
       flake = {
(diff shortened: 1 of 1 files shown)


roberth (approved):

roberth:
> defaultPackage

That one's been deprecated, so you made the right call.



> I wonder if packages.default shouldn't be a string? but then it would need validation...

As far as I know it should be a derivation in `packages`. [`apps.<name>.program`](https://flake.parts/options.html#opt-flake.apps._name_._name_.program) can be set to a string though (or a package with `meta.mainProgram`).

PR #38, opened 2022-06-17
introduce pkgsets

zimbatm:
Bundle the packages from the inputs into a set, so that they become more
easily accessible. This helps encourage composition over overlays.

See also https://github.com/numtide/devshell/pull/209

### DIFF
--- a/modules/nixpkgs.nix
+++ b/modules/nixpkgs.nix
@@ -20,6 +20,9 @@
             (inputs'.nixpkgs or (throw "flake-parts: The flake does not have a `nixpkgs` input. Please add it, or set `perSystem._module.args.pkgs` yourself."))
             inputs'.nixpkgs.legacyPackages
         );
+        _module.args.pkgsets = lib.mkOptionDefault (
+          lib.mapAttrs (name: value: value.packages or value.legacyPackages or {}) inputs'
+        );
       };
     };
   };


roberth:
I like the idea, but we should be very conservative about opinionated choices in the core.
Perhaps you could put this in a new module `flake-parts.flakeModules.pkgsets`? That way a flake can declare that it uses this non-native concept with `imports = [ flake-parts.flakeModules.pkgsets ];`.

I've set a bad example with the `nixpkgs.nix` module :(, so https://github.com/hercules-ci/flake-parts/issues/41

Do you think this module argument should only contain input package sets, or also user-provided ones, or module-provided ones (such as https://github.com/srid/haskell-flake)?

modules/nixpkgs.nix
```diff
@@ -20,6 +20,9 @@
             (inputs'.nixpkgs or (throw "flake-parts: The flake does not have a `nixpkgs` input. Please add it, or set `perSystem._module.args.pkgs` yourself."))
             inputs'.nixpkgs.legacyPackages
         );
+        _module.args.pkgsets = lib.mkOptionDefault (
```
roberth, line 23: Putting `mkOptionDefault` around the whole thing will cause the whole thing to disappear when the user (or a module) defines a pkgset of their own.
By moving it into the individual attrs, they become individualy overridable, but removing pkgsets will require a `mkForce` on the whole thing. This seems to be a better trade-off because removing isn't super useful unless someone has some radical ideas that I'm unaware of :)

modules/nixpkgs.nix
```diff
@@ -20,6 +20,9 @@
...
+          lib.mapAttrs (name: value: value.packages or value.legacyPackages or {}) inputs'
+        );
```
roberth, line 25:
```suggestion
        _module.args.pkgsets =
          lib.mapAttrs (name: value:  lib.mkOptionDefault (value.packages or value.legacyPackages or {})) inputs';
```

roberth:
> when passing flakes to NixOS or other module-based systems, the user must properly key the packages by the current system

There's a `withSystem` function in the top-level module arguments that may be useful.

> Also, it's useful to get access to `lib` and `nixosModules` from the flake.

Maybe you're already aware, but I'm trying to promote the fact that `specialArgs.flake = self`, etc are an antipattern.
`flake` is not a property of the object that the module system configures, so you'll run into trouble when multiple flakes contribute to that configuration, e.g. by using the `nixosModules` of a dependency.

The correct way to do it is _directly_ via the lexical scope. This means that the module that creates the link between the flake and the configuration is an anonymous module with free variables. (example free variable: `bar in `{ services.foo.package = bar; }`)

Putting this together we get something like

```nix
{
  nixosModules.foo = { pkgs, ... }: {
    services.foo.package = withSystem pkgs.hostPlatform.system ({ config, ... }: config.packages.foo);
  };
}
```

If we can establish a standard way for modules to find the "current system", we can create a new helper function `withSystemModule`, so that we can write

```nix
{
  nixosModules.foo = withSystemModule (
    perSystem@{ config, ... }:
    nixos@{ config, ... }: {
      services.foo.package = perSystem.config.packages.foo;
      services.bar.package = perSystem.config.packages.bar;
      imports = [ ./nixos-foo.nix ./nixos-bar.nix ];
    }
  );
}
```



> The second issue is that ideally, I am trying to make this work without flakes.

It's possible to import modules at the `perSystem` level, so with some care such non-flake perSystem-like modules may be made to interoperate. It seems to be rare for `perSystem` modules to access top-level flake bits. I suppose `inputs'` is a little awkward, so we might want to introduce a new module argument to unify these module system applications after all.

roberth:
Adding `withSystemModule` above as `moduleWithSystem` in #48. Does this help with your problem?

> Also, it's useful to get access to `lib` and `nixosModules` from the flake.

These are readily accessible in the `self` module parameter. I don't think there's much help we can offer, short of putting the flake in the config, which is bad habit that can't be used when the flake publishes the (NixOS, etc) module.

PR #45, opened 2022-06-27
Update type of `perSystem.packages` to allow for function and attrs

### LINKED ISSUE #44: Allow for more more than just packages in `perSystem.packages`
https://github.com/hercules-ci/flake-parts/blob/main/modules/packages.nix#L31 there the type of `perSystem.packages` is restricted to `lazyAttrsOf package`, which mostly is fine, but just in nixpkgs we have examples which won't fit there, namely `generators` or `writeText`. In my flake I define custom helpers of both the `attrsetOf (attrsOf package)` and `attrsetOf function` nature. I propose a new type:
```nix
let
  self = 
    oneOf
      [ (lazyAttrsOf self)
        package
        (functionTo self)
      ];
in
  lazyAttrsOf self // { description = "Attrs of functions or packges, with arbitrary depth."; }
```

MagicRB:
Signed-off-by: Magic_RB <magic_rb@redalder.org>

closes #44

### DIFF
--- a/modules/packages.nix
+++ b/modules/packages.nix
@@ -4,19 +4,30 @@ let
     filterAttrs
     mapAttrs
     mkOption
-    optionalAttrs
     types
     ;
   inherit (flake-parts-lib)
     mkSubmoduleOptions
     mkPerSystemOption
     ;
+
+  packageType =
+    with types;
+    let
+      self =
+        (oneOf
+          [ package
+            (lazyAttrsOf self)
+            (functionTo self)
+          ]) // { description = "Attrs of functions or packges, with arbitrary depth."; };
+    in
+      lazyAttrsOf self;
 in
 {
   options = {
     flake = mkSubmoduleOptions {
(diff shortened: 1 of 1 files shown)


MagicRB:
on hold

roberth:
I'm a big fan of nested attrsets, as they're convenient and lazier than a flattened attrset, but, unless there's some good news I'm unaware of, this is not supported by the flake schema.

```
$ nix flake check
error: flake attribute 'packages.x86_64-linux.foo' is not a derivation

$ nix --version
nix (Nix) 2.9.1
```

Nested attrsets _are_ supported in `legacyPackages`.

Functions are different, I'd say they should go in a `lib` attribute, although that's not officially part of the schema.
That'd be `flake.lib` in `flake-parts`.

I think we could explore an optional module to make defining a `lib` easier.

roberth:
Let's close this until Nix might change.

PR #46, opened 2022-07-06
templates: Use flake registry for flake-parts

roberth:


srid:
How does this work?

roberth:
Nix queries the flake registry (https://github.com/NixOS/flake-registry) for such "indirect" references and records the details in the lock file.

PR #56, opened 2022-09-19
add formatter schema

zimbatm:
This is needed for flakes that want to support `nix fmt`.

Right now it's missing some testing as I didn't find a test framework for it.

Needed by https://github.com/numtide/treefmt/pull/169

### DIFF
--- a/modules/formatter.nix
+++ b/modules/formatter.nix
@@ -28,7 +28,7 @@ in
       _file = ./formatter.nix;
       options = {
         packages = mkOption {
-          type = types.package;
+          type = types.nullOr types.package;
           default = null;
           description = ''
             A package used by <literal>nix fmt</literal>.
--- a/modules/formatter.nix
+++ b/modules/formatter.nix
@@ -1,0 +1,55 @@
+{ config, lib, flake-parts-lib, ... }:
+let
+  inherit (lib)
+    filterAttrs
+    mapAttrs
+    mkOption
+    optionalAttrs
+    types
+    ;
+  inherit (flake-parts-lib)
+    mkSubmoduleOptions
+    mkPerSystemOption
+    ;
+in
+{
+  options = {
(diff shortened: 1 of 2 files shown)


modules/formatter.nix
```diff
@@ -0,0 +1,55 @@
...
+        description = ''
+          Per system package used by <literal>nix fmt</literal>.
+        '';
+      };
+    };
+
+    perSystem = mkPerSystemOption ({ config, ... }: {
+      _file = ./formatter.nix;
+      options = {
+        packages = mkOption {
```
roberth, line 30:
```suggestion
        formatter = mkOption {
```

roberth:
Seems to work after suggestions and looks alright.

> Right now it's missing some testing as I didn't find a test framework for it.

Yeah the testing situation is kinda bad because I don't want to pollute the repo with test dependencies and such. People complain about those, even in other repos that aren't low level like this one. Not sure if I should wait for something like https://github.com/NixOS/nix/issues/6124

zimbatm:
> Yeah the testing situation is kinda bad because I don't want to pollute the repo with test dependencies and such. People complain about those, even in other repos that aren't low level like this one. Not sure if I should wait for something like https://github.com/NixOS/nix/issues/6124

Yeah, I know what you mean. One workaround would be to add a tests/flake.nix that refers back to the root flake. `nix run ./tests#.`?

roberth:
> One workaround would be to add a tests/flake.nix that refers back to the root flake. `nix run ./tests#.`?

Yeah, I'm doing something similar in the `dev` directory, though that's a very unusual not-quite-flake because you can't reference up (`../`), at least not when I wrote that solution.

zimbatm:
Oh well :) I think this is ready. I switched the upstream PR to this branch, and it's working as intended.

roberth (approved):

PR #58, opened 2022-09-28
Remove redundant null filters

### LINKED ISSUE #57: `perSystem` seemingly breaks with IFD
I essentially have `perSystem.packages = import drv`.
This doesn't work, and something tries to evaluate what `packages` is for all systems, even though it can only be evaluated for the current system.

roberth:
Fixes #57

### DIFF
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -83,11 +83,8 @@ in
   config = {
     flake.apps =
       mapAttrs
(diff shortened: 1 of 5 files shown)


L-as:
I just made #59 ! Maybe you want to merge mine anyways because I fixed the filtering.

L-as:
Thank you BTW.

PR #60, opened 2022-10-11
Only depend on nixpkgs/lib.

shlevy:
This reflects the actual requirements of the core flake-parts functionality and removes any possible confusion about where lib comes from (always flake-parts's input) vs where pkgs comes from (always the flake's nixpkgs input). Flakes which use flake-parts should only ever have to override its input in the very unlikely event of a bug in lib.

Hopefully lib will be separated into its own flake some day, which will also make this a much smaller footprint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,6 @@ Otherwise, add the input,
 
 ```
     flake-parts.url = "github:hercules-ci/flake-parts";
-    flake-parts.inputs.nixpkgs.follows = "nixpkgs";
 ```
 
 then slide `mkFlake` between your outputs function head and body,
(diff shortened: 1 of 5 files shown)


shlevy:
@roberth Oops, yes.

roberth (approved):
I don't think this change represents the only good way to depend on `lib` for flake parts, as both styles are acceptable and each have their pros and cons.
Nonetheless, this is a good change.

roberth:
Added to changelog https://github.com/hercules-ci/flake-parts/blob/main/ChangeLog.md#2022-10-11

roberth:
Thanks @shlevy

PR #61, opened 2022-10-16
Add option for defining flake-parts modules for downstream flakes.

shlevy:


srid:
What about `flake.flakeModules` instead?

Example: https://github.com/srid/emanote-template/blob/966f6240e2fbef3c4ff6c26aa222d2647aa98bfe/flake.nix#L14

shlevy:
Fine by me, defer to project maintainers.

ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+# 2022-10-16
+
+ - Added a new `flake.flakePartsModules` option so a flake can expose a module
+   to be used in a downstream flake's flake-parts usage.
```
roberth, line 4: `foo.flakePartsModules.default` is going to be quite a mouthful compared to the current `foo.flakeModule`.
On the other hand, `flakeModule` is a holdover from when this project was called `flake-modules`.

modules/flakePartsModules.nix
```diff
@@ -0,0 +1,27 @@
...
+    ;
+  inherit (flake-parts-lib)
+    mkSubmoduleOptions
+    ;
+in
+{
+  options = {
+    flake = mkSubmoduleOptions {
+      flakePartsModules = mkOption {
+        type = types.lazyAttrsOf types.unspecified;
```
roberth, line 18:
```suggestion
        type = types.lazyAttrsOf types.deferredModule;
```

This will add more detailed info about the individual definition(s), especially when the module is in its own file.

It's ok to keep the `apply` solution as well, because it creates a more stable name to use in `disabledModules`.

modules/flakePartsModules.nix
```diff
@@ -0,0 +1,27 @@
...
+        default = { };
+        apply = mapAttrs (k: v: { _file = "${toString self.outPath}/flake.nix#flakePartsModules.${k}"; imports = [ v ]; });
+        description = ''
+          flake-parts modules.
```
roberth, line 22:
```suggestion
          flake-parts modules for use by other flakes.
          
          If the flake defines only one module, it should be flakePartsModules.default.
          
          You can not use this option in the defining flake's imports. Instead you could create file for the module and reference it both in `imports` and in this option.
```

ChangeLog.md
roberth, line 4:
```suggestion
 - Added a new `flake.flakePartsModules` option so a flake can expose a module
   to be used in a downstream flake's flake-parts usage. This will replace
   `.flakeModule` in the future.
```

roberth (commented):
Renaming to `flakePartsModules` _seems_ like the right thing to do, even if I don't like the longer name. That's something that can be solved with a module system feature later.
 - https://github.com/NixOS/nixpkgs/issues/196584

ChangeLog.md
srid: I didn't know about the holdover. I had always assumed the name `flakeModule` to be appropriate because ... well .. flake-parts is the _only_ known module system (afaict) for flakes. And that, if/when flake-parts get subsumed into nixpkgs proper we would then have an official module system with the added hope of inheriting the `flakeModule` naming *as is*.

modules/flakePartsModules.nix
shlevy: Should this change be applied to nixosModules and darwinModules as well? I just copied those.

shlevy:
FYI let me know once the code looks good and I'll `git rebase --autosquash` down the fixup commits, I just like to use them in PRs that have already been reviewed so reviewers know what's changed and what's stayed the same.

terlar:
I was testing this and I am having some issues. I am not sure my issues come from this or from some other part of flake-parts.

If I have a flake (flake-parts-module-provider) that uses `flake-parts` and then expose a `flakePartsModule` via `flake.flakePartsModule.default`. Then use this module in another flake that uses `flake-parts` in an `imports` via `self.inputs.flake-parts-module-provider` you will get infinite recursion.

```
error: infinite recursion encountered

       at /nix/store/ircpgavsl5v9pipq6f23f1dzmlv8k9q0-source/lib.nix:72:32:

           71|       lib.evalModules {
           72|         specialArgs = { inherit self flake-parts-lib; inherit (self) inputs; } // specialArgs;
             |                                ^
           73|         modules = [ ./all-modules.nix module ];
```

Now this comes with using any `self` or `inputs` within `imports` as I understand it. It just becomes more apparent with this change as it is more likely that someone will use flake-parts to define the module with this and then consume the module with another flake-parts enabled flake.

I found the work-around to pass in the flake containing the `flakePartsModules` explicitly via `specialArgs`:

```
# ...
   outputs = {
    self,
    flake-parts,
    flake-parts-module-provider,
    ...
  }:
    flake-parts.lib.mkFlake {
      inherit self;
      specialArgs = {inherit flake-parts-module-provider;};
    } {
      systems = ["x86_64-linux" "x86_64-darwin" "aarch64-darwin"];
      imports = [./nix];
    };
}
```

I'm quite new to flake-parts so please forgive my ignorance if I have missed some essential information provided in the docs etc.

terlar:
Another case that I haven't found a work-around is that if you want to dog-food your own `flakePartsModules` in the flake providing the `flakePartsModules` you reach the same kind of error.

The work-around there is import the file directly instead. But would be nice if I could define `flakePartsModules.x` and then dog-food that same module in the same flake e.g. `imports = [self.flakePartsModules.x]`.

But perhaps that is asking for too much trouble?

roberth:
@terlar This is orthogonal to this PR. The same would happen with the current convention, `flake.flakeModule`.

> explicitly via `specialArgs`:

This adds an unnecessary indirection and it bloats the specialArgs. I'd import the modules in `flake.nix` itself, where the inputs are simply in scope. If you insist on importing in a separate module, you could use

```
{ self, ... }: {
  imports = [inputs.flake-parts-module-provider.flakePartsModules.default];
  # ...
}
```

> forgive my ignorance if I have missed some essential information provided in the docs etc.

We need more "guide" docs. Known problem. You've been forgiven ;)

> [error when you] dog-food your own `flakePartsModules`

This fundamentally can't be done from inside the fixpoint. You could define the module in a separate file or in a let binding to achieve this.

roberth:
@shlevy I find `.flakePartsModules.default` in the imports too unpleasant of a syntax. I liked the idea of improving the data model and make it consistent with the other module attributes, but I don't see a real user experience improvement compared to `.flakeModule`, especially as I haven't seen the need for a flake with multiple distinct flake modules yet.

roberth:
Taking one piece of the other thread here.

> > @shlevy I find `.flakePartsModules.default` in the imports too unpleasant of a syntax. I liked the idea of improving the data model and make it consistent with the other module attributes, but I don't see a real user experience improvement compared to `.flakeModule`, especially as I haven't seen the need for a flake with multiple distinct flake modules yet.

> ironically enough a policy that you just [disagreed with](https://github.com/hercules-ci/flake-parts/pull/61#issuecomment-1294954964) in one particular instance

I dislike the verbosity of the end user syntax of accessing the data model, not necessarily this part of the flakes data model. Does that explain the misunderstanding or did I miss some other irony?

shlevy:
@roberth I'm confused now. Do you basically agree with my 4 answers in the comment above then?

shlevy:
@roberth Updated according to my thinking in https://github.com/hercules-ci/flake-parts/pull/61#issuecomment-1295094542

Note that description for the alias doesn't render in the docs because the dev flake is not up-to-date with the new mddoc stuff.

ChangeLog.md
shlevy: Switched to `flake.flakeModules` and added `flakeModule` alias.

shlevy:
@roberth can this move forward?

roberth:
The change is not bad, but not without risk.

Good
 - adds documentation
 - defining flakeModules.<name> in multiple places (questionable)

Bad
 - this adds a flake output, unconditionally. Users already complain about useless attributes appearing in the repl and `nix flake show`
   - might be resolved by filtering out the non-transposed top-level attributes in a postprocessing step after the module system, such as `flake.flakeModules` and `flake.nixosConfigurations`, but this makes `self` more strict and therefore more susceptible to infinite recursions

Status quo: users can already define these attributes themselves

Maybe this should be an optional module, to be explicitly imported by flakes that want to expose a flake module?

shlevy:
@roberth Done

shlevy:
> but this makes self more strict and therefore more susceptible to infinite recursions

flake-parts users can always just access `config` for this use anyway, right? This is somewhat orthogonal to this PR but would be nice if e.g. `packages.x86_64-linux` were simply not exported if no packages are defined.

flake.nix
```diff
@@ -7,6 +7,11 @@
 
   outputs = { self, nixpkgs-lib, ... }: {
     lib = import ./lib.nix { inherit (nixpkgs-lib) lib; };
+    flakeModules = {
+      flakeModules = ./modules/flakeModules.nix;
+      default = self.flakeModules.flakeModules;
+    };
+    flakeModule = self.flakeModules.default;
```
roberth, line 14:
```suggestion
      flakeModules = ./modules/flakeModules.nix;
    };
```

I think the default is ambiguous in this case: should it be the core module(s) or the `flakeModules` module? Let's keep it simple.

roberth (commented):
One small change and a rebase :+1:

shlevy:
@roberth OK, ready to go!

roberth (approved):

roberth:
Thanks!

bors r+

PR #64, opened 2022-10-30
Add mkTransposedPerSystemModule for simple per-system exposed attrs

shlevy:


shlevy:
Note that docs won't properly render until flake-parts's doc generation is up to date with the mdDoc changes.

shlevy:
@roberth Any thoughts on this?

lib.nix
```diff
@@ -98,6 +98,34 @@ let
         type = flake-parts-lib.mkPerSystemType module;
       };
 
+    # Helper function for defining a per-system option that
+    # gets transposed by the usual flake system logic to a
+    # top-level flake attribute.
+    mkPerSystemModule = { name, option, file }: {
```
roberth, line 104:
```suggestion
    mkTransposedPerSystemModule = { name, option, file }: {
```

The transposition is quite relevant, as a perSystem module could be just one of those submodules.

It's also about a single attribute, but representing that seems to get too long.

dev/default.nix
```diff
@@ -1,6 +1,6 @@
 let
   flake = builtins.getFlake (toString ./.);
-  fmc-lib = import ../lib.nix { inherit (flake.inputs.nixpkgs) lib; };
+  fmc-lib = (builtins.getFlake (toString ../.)).lib;
```
roberth, line 3: `..` started out as a lockless flake. (fwiw)

This looks ok.

roberth (commented):
I didn't get around to this one, but it's not too complicated :)

lib.nix
shlevy: Done.

roberth (approved):

PR #80, opened 2022-11-27
Darwin cleanup

roberth:
Remove the `darwinModules` option. This was added in the early days
without full consideration. The removal will have no effect on most flakes
  considering that the [`flake` option](https://flake.parts/options/flake-parts.html#opt-flake)
  allows any attribute to be set. This attribute and related attributes should
  be added to the nix-darwin project instead.


ChangeLog.md
```diff
@@ -1,4 +1,12 @@
 
+# 2022-11-27
```
roberth, line 2:
```suggestion
# 2022-12-07
```

PR #82, opened 2022-12-01
Easy overlay publishing

roberth:
The goal here is to make a half-decent overlay easy to define, by overloading `pkgs` in `perSystem`.
Multiple kinds of overlays exist in flakes. This may not be the definitive way to define them, so it's not part of the core modules.

 - [ ] documentation

Example:

```nix
# ...
{
      imports = [
        inputs.flake-parts.flakeModules.easyOverlay
      ];
      perSystem = { config, pkgs, final, ... }: {
        overlayAttrs = { inherit (config.packages) my-package; };
        # ...
```

### DIFF
--- a/dev/tests/eval-tests.nix
+++ b/dev/tests/eval-tests.nix
@@ -1,5 +1,10 @@
 rec {
   f-p = builtins.getFlake (toString ../..);
+  flake-parts = f-p;
+
+  devFlake = builtins.getFlake (toString ../.);
+  nixpkgs = devFlake.inputs.nixpkgs;
+
   f-p-lib = f-p.lib;
 
   inherit (f-p-lib) mkFlake;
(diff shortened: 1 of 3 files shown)


srid:
Tangential thought: could this sort of API be useful for `haskell-flake` in some way? cf. https://github.com/srid/haskell-flake/issues/12

roberth:
> Tangential thought: could this sort of API be useful for `haskell-flake` in some way? cf. [srid/haskell-flake#12](https://github.com/srid/haskell-flake/issues/12)

Responded there. I think a module would be easier to understand, and more capable.

In general I think we should avoid overlays where we can use modules instead. Combining the two concepts is tricky.

shlevy:
This is really cool!

PR #87, opened 2022-12-20
Add debug module

roberth:
Whether to add the attributes `debug`, `allSystems` and `currentSystem`
to the flake output. When `true`, this allows inspection of options via
`nix repl`.

```
$ nix repl
nix-repl> :lf .
nix-repl> currentSystem._module.args.pkgs.hello
«derivation /nix/store/7vf0d0j7majv1ch1xymdylyql80cn5fp-hello-2.12.1.drv»
```

Each of `debug`, `allSystems.<system>` and `currentSystem` is an
attribute set consisting of the `config` attributes, plus the extra
attributes `_module`, `config`, `options`, `extendModules`. So note that
these are not part of the `config` parameter, but are merged in for
debugging convenience.
 - `debug`: The top-level options
 - `allSystems`: The `perSystem` submodule applied to the configured `systems`.
 - `currentSystem`: Shortcut into `allSystems`. Only available in impure mode.
           Works for arbitrary system values.

### DIFF
--- a/all-modules.nix
+++ b/all-modules.nix
@@ -3,6 +3,7 @@
   imports = [
     ./modules/apps.nix
     ./modules/checks.nix
+    ./modules/debug.nix
     ./modules/devShells.nix
     ./modules/flake.nix
     ./modules/formatter.nix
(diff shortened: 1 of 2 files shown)


cor (approved):
This looks like it is exactly what we needed! Thank you @roberth (CC:
@dzmitry-lahoda )

PR #90, opened 2022-12-24
Shrink inputs footprint further

blaggacao:
https://github.com/hercules-ci/flake-parts/commit/13dddfdc6701f05513fa77928ee90ad628b607c5
reads:

> Hopefully lib will be separated into its own flake some day, which
> will also make this a much smaller footprint.

Here it is.

/cc @shlevy

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,19 +2,16 @@
   "nodes": {
     "nixpkgs-lib": {
       "locked": {
-        "dir": "lib",
-        "lastModified": 1671359686,
-        "narHash": "sha256-3MpC6yZo+Xn9cPordGz2/ii6IJpP2n8LE8e/ebUXLrs=",
-        "owner": "NixOS",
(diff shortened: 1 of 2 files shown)


roberth:
In theory, a github git subtree should be no less efficient than whole repo of equal size.
Nix should implement subdir fetching with the github api as follows. Crucially, the github archive endpoint supports tree hashes.

At lock time
 - look up the tree for a given commit
 - store the git subtree sha
 - fetch the archive for the git subtree sha

At fetch time
 - ignore rev, and fetch the subtree archive

zimbatm:
I think this can be closed since now flake.parts uses `github:NixOS/nixpkgs/nixos-unstable?dir=lib` as the input

blaggacao:
It doesn't mean though the goal of this PR is already achieved since I infer from Robert's comment that this would require upstream changes to Nix.

PR #103, opened 2023-01-10
add lib argument to multi-module template

DavHau:
Might be biased by my use cases but I nearly always need lib. Even if not it doesn't hurt to have it.
Also the top-level flake.nix could really use a `lib` attribute. But not sure what would be the best way to bring it in scope.
Many users will need lib, so I think it would be good to propose a sane default of how to bring it in.


roberth (approved):

roberth:
Looks good, thanks!

bors r+

Issue #51, opened 2022-08-12
Idiomatic way to provide a "library" of Nix functions


srid:
We are using the module system options to expose functions for other modules to use.

For example,

```nix
{ self, ... }:
{
  options = {
    mkCheck = lib.mkOption {
      type = lib.types.functionTo lib.types.unspecified;
      default = { arg1, arg2 }: 
        abort "Not implemented";
    };
  };
}
```

Then we use it from other flake-module.nix as:

```nix
{ self, ... }: {
  perSystem = { config, ... }: {
    checks.foo = config.mkCheck { ... };
  };
}
```

Is this the recommended way to use the module system to achieve this?

roberth:
I don't think there's an idiomatic way yet, but we can think of a good opinionated way to do it.

Some requirements:
 - good namespacing
 - consider both libraries for internal use and for external use
 - consider libraries that depend on `system` and those that don't
    - `system`-dependent ones shouldn't have to pass `system`
      - refs #48
    - `lib` flake attribute should be an attrset, but may contain a `system`-dependent function for constructing a `system`-dependent library
 - any generic support modules should go in `flakeModules.<foo>` or a separate repo, not the default module list, which is supposed to be unopinionated, capturing only what `nix` itself hardcodes

Possible design:

Defining and using public libs (`<flake>.lib`)
 - Use `flake.lib` for putting lib attributes that don't necessarily depend on `system`
 - Use `perSystem.lib` for lib attributes that do depend on `system`
 - Expose `perSystem.lib` in `flake.lib.libWithSystem = system: ...`
 - Define `inputs'.<name>.lib` to be `inputs.<name>.lib // inputs.<name>.lib.libWithSystem system` (or `// {}` if not exists)
 - Define `perSystem` module argument `libs` so that `libs.<name>` = `inputs'.foo.lib`

Defining a lib to be used in a `<flake>.flakeModule` (module to be used in another flake)
 - Option a. Define a `<flake>.lib`, define `flakeModule` with `withSystem` or #48 
   - This lib being public is probably a good thing?
 - Option b. Something else? We can't rely on `inputs'` because the consumer flake should be free to have the flake inputs it wants with the names it wants.

Defining a lib to be used locally
 - Similar to previous, but `libs.self` may be ok for non-public flake module. (Similar to how `specialArgs` should only be used by the final consumer, e.g. `<flake>.nixosConfigurations`, for this consumer's own purposes only)
[...]

PR #109, opened 2023-01-29
Add the ability to pass lib overlays to mkFlake

clhodapp:
Partially address https://github.com/hercules-ci/flake-parts/issues/51 by providing a way to specify a list or attribute set of library overlays when you invoke `mkFlake`. These have to be passed at the point where `mkFlake` is called because we need to compose the new `lib` before we call `evalModules` (since we need to call `evalModules` on a version of `lib` that has the overlays applied).

Documentation in https://github.com/hercules-ci/flake.parts-website/pull/116

This does _not_ address per-system `lib` attributes at this time.

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -85,7 +85,66 @@ let
     arguments as `inputs` like above.
   '';
 
+  # Use a list of lib overlays to extend a lib
+  mkExtendedLib = (
+    let libExtend = lib: lib.extend;
+    in overlays: lib: builtins.foldl' libExtend lib overlays
+  );
+
+  # The overlays we are passed could be any of:
+  #
+  # (1) A simple attribute attribute set
+  # (2) A function from `prev` to an attribute set
+  # (3) A function from `final` and `prev` to an attribute set
+  # (4) A function from `inputs`, `final`, `prev`, to an attribute set
+  # (5) A path or string pointing at a file containing any of (1) - (4)
+  #
+  # We seek to normalize these possibilities so that we always have a function
+  # from `final` and `prev` to an attribute set, which is suitable for use with
+  # mkExtendedLib. This may mean specializing on the `inputs` that we are passed.
+  #
+  toStandardOverlay = inputs: freeformOverlay: (
+    let
+      reifiedFreeformOverlay = (
+        if (builtins.isPath freeformOverlay) || (builtins.isString freeformOverlay)
+        then import freeformOverlay
+        else freeformOverlay
+      );
+    in (final: prev:
+      # apply the correct number of arguments to `reifiedFreeformOverlay`. Of note
+      # is the fact that the bindings that pass too many args should not blow up
+      # because laziness and short-circuiting should ensure that they are never
+      # actually invoked.
+      let
+        noArg = reifiedFreeformOverlay;
+        oneArg = reifiedFreeformOverlay prev;
+        twoArg = reifiedFreeformOverlay final prev;
+        threeArg = reifiedFreeformOverlay inputs final prev;
+      in
+        if ! builtins.isFunction noArg then noArg
+        else if ! builtins.isFunction oneArg then oneArg
+        else if ! builtins.isFunction twoArg then twoArg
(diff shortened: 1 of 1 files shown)


clhodapp:
Note: It's possible that basic `lib` overlay functionality belongs a level down in `evalModules` but I suspect that there'd need to be _quite_ a lot of discussion to get it merged into `nixpkgs`.

roberth:
I don't think `lib` was meant to be an extension point, despite `lib.extend`. It allows accidental monkey-patching and the pattern requires all extensions to be added more or less simultanously, and before the flake-parts fixpoint gets to run. As such, the module system isn't of any help either.


Instead, I think we could use the `lib` attribute from the inputs (e.g. `inputs.nixpkgs.lib`, `inputs.self.lib`), which you can already do today, and perhaps try to introduce a convention for system-dependent libraries, or perhaps even `pkgs`-dependent libraries.

clhodapp:
Yeah, I suspect that lib overlays becoming "a thing" is likely to be controversial for folks interested in engineering the nix ecosystem because it may not be necessary and it feels like opening a door that can't be closed. The truth is, though, this pattern is quietly already "a thing". In fact it's super-common, especially with flakes. Here's a small selection of proof that this is the case:

* home-manager: https://github.com/nix-community/home-manager/blob/4fa1ba72a39da509399cce43995e4c1e1960ce9c/modules/lib/stdlib-extended.nix#L7
* kube-nix: https://github.com/hall/kubenix/blob/5bcab30ed204bfbf8ff6a37e67af154ff8aac16e/flake.nix#L46
* nixpkgs itself: https://github.com/NixOS/nixpkgs/blob/8e7ecb51e01f0204534d766e97b8207a1fd62d9f/flake.nix#L17
* Digga (née DevOS): https://github.com/divnix/digga/blob/54ede8e591d288c176a09d6fcf4b123896c0bf0f/src/mkFlake/fup-adapter.nix#L22
* Lots and lots of people in their personal nix configs https://github.com/search?q=%22lib.extend+%22++filename%3Aflake.nix&type=Code

Note that GitHub search is pretty fuzzy so, while a vast majority of the search results are folks extending `lib` in their `flake.nix`, a few of them are other patterns, such as calling `extendDerivation`.

I would argue that there is essentially no difference in benefits and drawbacks between library overlays and package overlays. You said, "It allows accidental monkey-patching and the pattern requires all extensions to be added more or less simultanously...". This exactly applies to package overlays just as much as library overlays and yet we all use package overlays heavily because they're so useful. Personally, I follow this advice for both my package overlays and my library overlays: https://github.com/hercules-ci/flake.parts-website/pull/116/files#diff-7b7db4f9481c5eec5efff2c5981b816afad8e70395c09561f96f1f6931e33c1eR31-R38. So far, it has prevented me for having any serious composition issues.

You note that `lib` composition must happen "before the flake-parts fixpoint gets to run." That's actually the exact reason for this patch!  I would strongly have preferred to do something more through the module system. However, that's simply not possible with the way that module composition currently works. This is what I was getting at with my comment about this potentially going a level down in `evalModules`. Perhaps something more sophisticated could be done with respect to module staging to allow something that's logically in between `specialArgs` and `args` to exist inside of the module system.

One thing to note: I had thought that I'd done the research/experiments to conclude that `evalModules` was coded to _always_ pass the `lib` it was called on as the `lib` argument to modules. However, in the course of writing this comment, though, I discovered that you can actually override the `lib` argument by passing a `lib` special arg.


With respect to depending on `lib` attributes on your `inputs`: I have seen this as a recurring suggested pattern. However, my experience is that this works OK for flake-parts modules but it's rather annoying/cumbersome when you start doing nested module compositions for things like home-manager or nixos. It's absolutely _possible_ to make `inputs` available at this level, of course, but it gets pretty messy.


All this said: I don't technically _need_ this change (there are other ways I can accomplish what I'm going for) so I'll close the PR's if you're still not convinced. Let me know what you think!

clhodapp:
I'm thinking I'll just close this.

I can make it anew if the case for lib overlays ever gets stronger.

Thanks!

PR #118, opened 2023-02-09
Fix highlighting in README.md

IogaMaster:
Just a little aesthetic change.


roberth:
I didn't do it because GitHub doesn't like the `=` and `;` in expression fragments that aren't exactly complete expressions, but I suppose that's a GitHub bug and not a README bug.

PR #126, opened 2023-03-21
Draft for nixpkgs flake module

maisiliym:
Uses https://github.com/NixOS/nixpkgs/pull/222155
This is shared currently to document and test.
Done from a [live coding on youtube](https://youtube.com/live/lRNPDW8xFoA)

Sajban-style-preview commit message:
``` shen
(added (naiveImplementationFor (nixpkgs flakeModule)))
```

### DIFF
--- a/modules/nixpkgs.nix
+++ b/modules/nixpkgs.nix
@@ -15,8 +15,10 @@ in
           # Last element has priority, so `perSystem` overlays rule
           (topLevel.config.nixpkgs.overlays ++ config.nixpkgs.overlays);
 
-        finalPkgs = config.nixpkgs.mkPkgsFromArgs
-          { inherit system; overlays = [ finalOverlay ]; };
+        finalPkgs = config.nixpkgs.mkPkgsFromArgs {
+          localSystem = { inherit system; };
+          overlays = [ finalOverlay ];
+        };
 
       in
       {
--- a/modules/nixpkgs.nix
+++ b/modules/nixpkgs.nix
@@ -1,13 +1,8 @@
-#
 # Nixpkgs module. The only exception to the rule.
-#
 # Provides a `pkgs` argument in `perSystem`.
-#
 topLevel@{ config, options, inputs, lib, flake-parts-lib, ... }:
 let
-  inherit (flake-parts-lib)
-    mkPerSystemOption
-    ;
+  inherit (flake-parts-lib) mkPerSystemOption;
 
 in
 {
--- a/modules/nixpkgs.nix
+++ b/modules/nixpkgs.nix
@@ -11,20 +11,24 @@
 # will be accepted into flake-parts, because it's against the
 # spirit of Flakes.
 #
-topLevel@{ config, options, inputs, lib, ... }:
+topLevel@{ config, options, inputs, lib, flake-parts-lib, ... }:
+let
+  inherit (flake-parts-lib)
+    mkPerSystemOption
+    ;
+in
 {
+  imports = [ inputs.nixpkgs.flakeModules.default ];
+
   options = {
     perSystem = mkPerSystemOption ({ config, system, ... }: {
       _file = ./nixpkgs.nix;
+      imports = [{ inherit (topLevel.config) nixpkgs; }];
       options = { inherit (options) nixpkgs; };
       config = {
-        imports = [{ inherit (topLevel.config) nixpkgs; }];
-        _module.args.pkgs = lib.mkDefault (config.allPkgsPerSystem.${system});
+        _module.args.pkgs = lib.mkDefault (config.nixpkgs.allPkgsPerSystem.${system});
       };
     });
   };
 
-  config = {
-    imports = [ inputs.nixpkgs.flakeModule ];
-  };
 }
--- a/modules/nixpkgs.nix
+++ b/modules/nixpkgs.nix
@@ -11,14 +11,20 @@
 # will be accepted into flake-parts, because it's against the
 # spirit of Flakes.
 #
-{ config, inputs, lib, ... }:
+topLevel@{ config, options, inputs, lib, ... }:
 {
+  options = {
+    perSystem = mkPerSystemOption ({ config, system, ... }: {
+      _file = ./nixpkgs.nix;
+      options = { inherit (options) nixpkgs; };
+      config = {
+        imports = [{ inherit (topLevel.config) nixpkgs; }];
+        _module.args.pkgs = lib.mkDefault (config.allPkgsPerSystem.${system});
+      };
+    });
+  };
+
   config = {
     imports = [ inputs.nixpkgs.flakeModule ];
-    perSystem = { ... }: {
-      config = {
-        _module.args.pkgs = lib.mkOptionDefault config.nixpkgs.output;
-      };
-    };
   };
 }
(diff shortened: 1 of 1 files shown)


modules/nixpkgs.nix
roberth, line 1: I don't think additions to this module are necessary.

modules/nixpkgs.nix
```diff
@@ -1,26 +1,32 @@
...
-    perSystem = { inputs', lib, ... }: {
-      config = {
-        _module.args.pkgs = lib.mkOptionDefault (
-          builtins.seq
-            (inputs'.nixpkgs or (throw "flake-parts: The flake does not have a `nixpkgs` input. Please add it, or set `perSystem._module.args.pkgs` yourself."))
-            inputs'.nixpkgs.legacyPackages
-        );
-      };
-    };
+  imports = [ inputs.nixpkgs.flakeModules.default ];
```
roberth, line 9: We shouldn't make assumptions about the `inputs`, especially not in the core of flake parts.

modules/nixpkgs.nix
```diff
@@ -1,26 +1,32 @@
...
+          localSystem = { inherit system; };
+          overlays = [ finalOverlay ];
+        };
+
+      in
+      {
+        imports = [ inputs.nixpkgs.flakeModules.default ];
+        _file = ./nixpkgs.nix;
+        config = {
+          _module.args.pkgs = lib.mkForce finalPkgs;
```
roberth, line 28: I would give this responsibility to `nixpkgs.flakeModules.default`, and leave the original default in place for the time being.

modules/nixpkgs.nix
```diff
@@ -1,26 +1,32 @@
...
+          (topLevel.config.nixpkgs.overlays ++ config.nixpkgs.overlays);
+
+        finalPkgs = config.nixpkgs.mkPkgsFromArgs {
+          localSystem = { inherit system; };
+          overlays = [ finalOverlay ];
+        };
+
+      in
+      {
+        imports = [ inputs.nixpkgs.flakeModules.default ];
```
roberth, line 25: As a general rule, modules are imported at the top level only. This makes importing more consistent, not having to wonder where to import it.
Maybe this was added by mistake? Maybe it doesn't even raise an error, as the laziness can mask problems sometimes; a blessing and a curse.

roberth (commented):
Hi Li, how's it going?

I think it should be possible to turn most if not all changes here into additions to the `nixpkgs.flakeModules.default` module. Perhaps we could add a deprecation message to the old default, suggesting that they import the `nixpkgs` module themselves. The idea is to reinforce the modular architecture and give users more choice.
What do you think?

roberth:
> I think it should be possible to turn most if not all changes here into additions to the `nixpkgs.flakeModules.default` module.

I see you've avoided the `perSystem` option there. I think users would expect the Nixpkgs to be in `perSystem`.
Maybe you've avoided it because that might not be allowed architecturally, but I don't think this should be a concern. It's very common for modules to make assumptions about the environment they're in, and for a flake-parts module that environment includes `perSystem`.
Another reason I suppose is that it would be nice to have a very pure module that's all about calling the Nixpkgs function and very little else. I think such a module would be great to have, and it would be probably be something that you could import into the `flakeModules.default` module, inside `perSystem`.
This creates a bit different division of labor, where it's also possible to for example have an overlay depend on an option that's in `perSystem`.

Here's what I think that might look like.

A general nixpkgs invocation module, not specific to any context (where flake-parts is such a context):

`pkgs/top-level/modules/pkgs.nix`
```nix
{ config, lib, ... }: {
  options.nixpkgs = {
    overlays = ...;
    system = ...; # or even hostPlatform + buildPlatform options
    config = ...;
  }
  config._module.args = lib.mkDefault (import .. { ... });
}
```

The module for `nixpkgs.flakeModules.default`:

`pkgs/top-level/modules/flakeModules.nix`
```
{
   options.perSystem = mkPerSystemOption ({ config, system, ... }: {
     imports = [ ./pkgs.nix ];
     nixpkgs.system = system;
   };
}
```

modules/nixpkgs.nix
LiGoldragon: I thought this would be needed to expose the `nixpkgs` option to the `perSystem` evaluation scope. Will experiment with removing it now.

modules/nixpkgs.nix
LiGoldragon: I thought this would be needed to expose the `nixpkgs` option to the `perSystem` evaluation scope. Will experiment with removing it now.

LiGoldragon: > I see you've avoided the perSystem option there. I think users would expect the Nixpkgs to be in perSystem.

See my comment above. Am I wrong? I'm using `nixpkgs` option inside the `perSystem` scope [here](https://github.com/sajban/clojix/blob/f1e58a7fee1ebcdc0dee82781d1ca81c95a9cfd1/nix/packagesModule.nix#L94)

LiGoldragon:
> This creates a bit different division of labor, where it's also possible to for example have an overlay depend on an option that's in perSystem.

Beat you to it: [clojix does that](https://github.com/sajban/clojix/blob/f1e58a7fee1ebcdc0dee82781d1ca81c95a9cfd1/nix/packagesModule.nix#L23)

LiGoldragon:
> `pkgs/top-level/modules/flakeModules.nix`
> 
> ```
> {
>    options.perSystem = mkPerSystemOption ({ config, system, ... }: {
>      imports = [ ./pkgs.nix ];
>      nixpkgs.system = system;
>    };
> }
> ```

That would make nixpkgs depent on flake-parts for `mkPerSystemOption`, right?

After reconsidering this from a few steps back, I realized none of this really needs to go in either flake-parts or nixpkgs. In fact, I'm rather of the opinion that nothing should go in nixpkgs anymore, since flakes is partly about splitting giant mono-repos. I have a rather funny story related to that; a few years ago when I began to play with flakes, I was going to make a flake for every major function, only to realize (after creating like 50 repos), that functions can't use caching very well, if at all.

So I might just create a repo and call it 'nixpkgs-flake' or something. Stay tuned.

And thanks for all the help. 

Off-topic: I would love to know your birth time and location, I would give you a quick astrology reading one of these days. Get the exact birth time if possible, to the minute. Just looking at your profile pic and from reading your comments and code, my first bet is Virgo. Did I get that right? Lol! Cheers! ;)

roberth:
> Off-topic: I would love to know

The universe is far more mysterious than that... :star2:

LiGoldragon:
I'll close as unattended, will make a new one if I take another shot, as it would be different now. Related to a module in nixpkgs, which should share some code as you said. Lib sounds right, since that is nixpkgs's lib, and overlays are one of its API types. I want to move away from overlays for sure though. I think a github-like namespace would do a lot, especially with pre-typed subnames: `hub.LiGoldragon.website` is a `website` type, allowing for safer jobs for shared servers, etc. I have a repo called `hob` where I'll be doing a complete rewrite in clojure soon. Stay tuned.

LiGoldragon:
> > Off-topic: I would love to know
> 
> The universe is far more mysterious than that... star2

I'm trying to see if that means Virgo or not ...
More mysterious than anyone can imagine in his wildest fantasy.

PR #132, opened 2023-04-03
Added `outPath` attribute to `self'` and `flakeModules` to `inputs'`

LiGoldragon:
Very straightforward. Was needed for [clojix](https://github.com/sajban/clojix/blob/daf44b5196071730bca5804578c177f2b4dc218a/nix/projectSubmodule.nix#L203).


roberth:
Both changes could perhaps be implemented more cleanly with an attribute such that `self'.flake == self`, as suggested in #152

LiGoldragon:
Yes, good points. Forgot I also was adding `flakeModules` afterwards. I'll be more careful with my PR branches! Thanks.

Issue #119, opened 2023-02-13
Dev-only dependencies


roberth:
Avoid loading dev-only modules by deferring those; when `checks` etc are accessed.
 - `devImports : deferredModule`
 - override `flake.checks` and `flake.devShells` in the main eval; choice of attrs should be generalized
 - what about apps?

roberth:
`flakeModules` may suffer from a similar lack of laziness and should perhaps be considered as a "class" or "context" of its own.

Both it and dev-only dependencies move towards a situation where the top level attributes are more controlled, and have to be enumerated somewhere in order to be published on the flake.
It's perhaps worth considering to model these dependency phenomena explicitly within the top level modules somehow. The genericity seems warranted because the internal dependencies may differ between flakes. I'm concerned about the complexity of such a solution though, and about the potential for bad interactions (although that might just be fear talking at this point).
To develop this, it's probably best to start with dev-only dependencies and perhaps later generalize it, if feasible.

roberth:
Here's a possible implementation
 - https://github.com/hercules-ci/hercules-ci-agent/pull/516

The flaws are
 - Compensating for bad subflake support in Nix, while the subflake is actually a workaround for https://github.com/NixOS/nix/issues/7730
 - Re-evaluating the modules without restriction could let the dev modules make changes to e.g. packages, potentially defeating `checks` / CI. While this is nothing new - you could test `mypkg.override foo` instead of `mypkg`, it does make it possible to override something by mistake.

Ideally it would reuse the base evaluation's option values, while also setting `readOnly = true` on the pre-existing options. Only specific options, such as `checks` and `perSystem.checks` would have their _definitions_ forwarded, allowing merging (instead of forwarding the final option value).

> a situation where the top level attributes are more controlled, and have to be enumerated somewhere in order to be published on the flake.

Not sure. They may be quite orthogonal despite serving similar goals.

>  later generalize it

Had no problem generalizing https://github.com/hercules-ci/hercules-ci-agent/pull/516 (mentioned at the start).

roberth:
:tada: You may now use [flake-parts.flakeModules.partitions](https://flake.parts/options/flake-parts-partitions) to achieve this.

-----

I'm debating whether to add an opinionated module that sets
```nix
partitionedAttrs.checks = "dev";
partitionedAttrs.devShells = "dev";
```
and perhaps add a convenience alias `dev` for `partitions.dev.module`.
[...]

PR #133, opened 2023-04-09
support externally extensible systems

zimbatm:
With this change, it becomes possible to modify the list of systems that
a flake is using, using only the flake override mechanisms.

See <https://github.com/numtide/treefmt/pull/228> and
<https://github.com/nix-systems/nix-systems>.

### DIFF
--- a/modules/perSystem.nix
+++ b/modules/perSystem.nix
@@ -59,14 +59,21 @@ let
 in
 {
   options = {
-    systems = mkOption {
+    systems = mkOption ({
       description = ''
         All the system types to enumerate in the flake output subattributes.
 
         In other words, all valid values for `system` in e.g. `packages.<system>.foo`.
+
+        If the `systems` flake input exists, it will be imported as the
+        default. See <https://github.com/nix-systems/nix-systems> for a
+        description of the pattern.
       '';
       type = types.listOf types.str;
-    };
+    } // (lib.optionalAttrs self.inputs?systems) {
+      # Load the systems input by default if it has been declared.
+      default = import self.inputs.systems;
+    });
 
     perInput = mkOption {
       description = ''
--- a/template/default/flake.nix
+++ b/template/default/flake.nix
@@ -3,6 +3,8 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    # The list of systems to use
+    systems.url = "github:nix-systems/default";
   };
 
   outputs = inputs@{ flake-parts, ... }:
@@ -14,7 +16,6 @@
         # 3. Add here: foo.flakeModule
 
       ];
-      systems = [ "x86_64-linux" "aarch64-darwin" ];
       perSystem = { config, self', inputs', pkgs, system, ... }: {
         # Per-system attributes can be defined here. The self' and inputs'
         # module parameters provide easy access to attributes of the same


modules/perSystem.nix
```diff
@@ -59,14 +59,21 @@ let
...
+        If the `systems` flake input exists, it will be imported as the
+        default. See <https://github.com/nix-systems/nix-systems> for a
+        description of the pattern.
       '';
       type = types.listOf types.str;
-    };
+    } // (lib.optionalAttrs self.inputs?systems) {
+      # Load the systems input by default if it has been declared.
+      default = import self.inputs.systems;
+    });
```
roberth, line 76: This doesn't really fit the architecture, because we shouldn't be making assumptions about the user's flake inputs; certainly not in the core, non-optional modules.

It'd be nicer to have nix-systems implement the flake-parts interface, by providing a module that just sets the `systems` option. In that case it doesn't need to be a default either.

zimbatm: Does flake-parts auto-import inputs' flakeModules?

Alternatively, I would propose extending the flake-parts "systems" type to support `inherits systems;` when systems is an input.

roberth: By default it does try to import all inputs, and I wouldn't encourage that, because that would defeat lazy fetching and any solution to #119 
Accepting a flake (I presume?) for an arbitrary option does not seem like a sensible solution to me. It abbreviates the syntax, but that is not a goal. An option should have a meaningful type.

bb010g: > This doesn't really fit the architecture, because we shouldn't be making assumptions about the user's flake inputs; certainly not in the core, non-optional modules.

This already happens with the Nixpkgs module's `_module.args.pkgs` assuming `inputs.nixpkgs`. This is justified through it being an important, special case. I do think the same could be said here.

> It'd be nicer to have nix-systems implement the flake-parts interface, by providing a module that just sets the `systems` option. In that case it doesn't need to be a default either.

`nix-systems/default` *can't* provide a module. All consumers have to assume that `inputs.systems.flake = false;` for the simple overrides this pattern allows to work properly. (`inputs.systems.url = "file:./systems.nix";` is very intentionally valid.)

I would recommend changing the fancy conditional attrset merge here with normal
```nix
mkOption {
  # …
  default = import inputs.systems or (throw "flake-parts: The flake does not have a `systems` input. Please add it, or set `systems` yourself.");
  defaultText = literalExpression ''import inputs.systems'';
}
```
like the Nixpkgs module does.

roberth: > This already happens with the Nixpkgs module's `_module.args.pkgs` assuming `inputs.nixpkgs`.

That's a temporary measure and a bad precedent, and documented as such in the code. I will remove it.

> `"file:./systems.nix"`

I wasn't aware of that requirement.

I think the standard way is good enough then. All this fancy `system` inheriting stuff is just a ridiculous workaround anyway.

```
systems = import inputs.systems;
```

bb010g: > I think the standard way is good enough then. All this fancy `system` inheriting stuff is just a ridiculous workaround anyway.
> 
> ```
> systems = import inputs.systems;
> ```

That's fair; the idea behind this would be to reduce boilerplate just a little bit more and push people towards just using `inputs.systems` instead of deciding to list the inputs out inline.

bb010g: > > This already happens with the Nixpkgs module's `_module.args.pkgs` assuming `inputs.nixpkgs`.
> 
> That's a temporary measure and a bad precedent, and documented as such in the code. I will remove it.

It's probably better to discuss this more in <https://github.com/hercules-ci/flake-parts/pull/137>, but I think it would be a shame to see that module go before a capable `inputs.nixpkgs.flakeModules.default` exists. Virtually every flake will need Nixpkgs, and having to manually set up `config._module.args.pkgs` boilerplate or import a 3rd-party flake module for it will push people away from using flake-parts when they're first evaluating it. At the very least, some sort of `inputs.nixpkgs-flake-parts.url = "github:hercules-ci/nixpkgs-flake-parts";` needs to exist first, before this is removed. I think the Nixpkgs module is an okay exception until flake-parts picks up more momentum.

zimbatm: @roberth what is your conclusion? I wouldn't mind re-opening the PR if it's possible to move forwards.

PR #145, opened 2023-05-07
Add mkDeferredModuleOption (fix #142)

### LINKED ISSUE #142: Better naming `mkPerSystemOption`
`mkPerSystemOption` barely creates an option of deferred module for documenting purpose. Shall we rename it to `mkDeferredModuleOption`? Since it is general-purposed, shall we upstream it to `nixpkgs`?

Atry:


roberth (approved):

roberth:
GitHub is not creating a queue...

bors r+

roberth:
Looks like a whitespace error. Could you fix that?

https://hercules-ci.com/accounts/github/hercules-ci/derivations/%2Fnix%2Fstore%2F1xm06xa1brz19ciki6g82bkh47lag05z-pre-commit-run.drv/log?via-job=89e4ac12-5fa2-4c94-a534-9cd79cf541b1

roberth:
Sorry about the bot mess; going to work on that.

roberth:
# Deprecated

Hey @Atry, I wouldn't recommend using these anymore, because the bundled deferred module type is subtly different from upstream Nixpkgs by returning a list of modules instead of a single module.

For now I've only documented a deprecation in #334. The `flake-parts-lib.mkDeferred*` names will start emitting a warning in the future.
That will be a preparation to switch `perSystem` itself to the upstream type that does not return a list, so `mkPerSystem*` are the correct names for type merging into `perSystem`. Since they have a clear purpose, I can reasonably change them to match the updated option. That does not hold true for the generalized names unfortunately.

PR #147, opened 2023-05-07
Add nixpkgs module (fix #74)

### LINKED ISSUE #74: Nixpkgs module
Prototype a more extensive Nixpkgs module in flake-parts before perhaps eventually doing #41 

https://twitter.com/ldesgoui/status/1592576468789628928

Atry:


extras/nixpkgs.nix
```diff
@@ -0,0 +1,8 @@
+{inputs, ...}: {
+  options.perSystem = mkPerSystemOption ({ pkgs, system, ... }: {
+    imports = [
+      "${inputs.nixpkgs}/nixos/modules/misc/nixpkgs.nix"
```
Atry, line 4: `inputs.nixpkgs` makes an assumption to the name of user's inputs, which is bad. We still need to solve #157 to properly use an input.

roberth:
Although I like the sort of elegance of such a short solution, I don't think we should be making the module list of `perSystem` strict in a Nixpkgs input.

For instance, when a flake exports an overlay with `easyOverlay`, `pkgs` will be set to the overlay `final` parameter, and it would be bad to have to fetch an unused Nixpkgs just to load its entrypoint module, and then find out that the only effect of that module is going to be overridden anyway.

So architecturally, it should look more like #137, except perhaps a simpler implementation.

PR #152, opened 2023-05-08
perSystem: include outPath in flake attributes

Pacman99:
I want to pass `self'` to my nixos modules and be able to do `"${self'}/...` to refer to paths within the flake. I think it would be good for flake-parts to preserve `outPath` within `self'` and inputs in `inputs'`, so that way `self'` and `inputs'` are identical to `self` and `inputs` but with system picked out.

### DIFF
--- a/modules/perSystem.nix
+++ b/modules/perSystem.nix
@@ -130,6 +130,11 @@ in
     # where we run in the context of an overlay, and the performance cost of the
     # extra `pkgs` makes the cost of running `perSystem` probably negligible.
     _module.args.getSystemIgnoreWarning = system: config.allSystems.${system} or (config.perSystem system);
+
+    perInput = system: flake: {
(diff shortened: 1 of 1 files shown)


roberth:
For context, this PR was discussed here in
 - https://github.com/hercules-ci/flake-parts/issues/153#issuecomment-1538975473

I think flake-parts should promote encapsulation, but you are free to define this addition in your own flake for your convenience. I.e.

```
{
  perInput = system: flake: {
    # Allow for input coercion to string, ex: "${inputs'.flake-parts}/shell.nix"
    inherit (flake) outPath;
  };
}
```

I'd be open to adding this little module as `flake-parts.flakeModules.perSystemOutPath`. (`perInputOutPath` is probably not as obvious to readers)

Another idea is to add a `perInput` attribute like `flake`, such that `self'.flake == self`; perhaps in `perSystem.nix`'s `config`.

PR #154, opened 2023-05-08
Add class: imports "type checking"

roberth:
Might delay a merge as it could lead to something like "unknown option _class" when loading modules from flakes with a (very) recent lib into flakes with an older lib.
Make a tag first?


cdmistman:
👋 @roberth - since #223 and #224 were closed in favor of the stack starting here, i was wondering what work still needs to be done to get this merged, and if i can help in any way?

cdmistman:
ack - haven't gotten to it yet, will get to it this weekend!

PR #155, opened 2023-05-08
Mention the trick to mkFlake twice

Atry:


shanesveller:
Is that approach really something the project should tacitly endorse by mentioning it in first-party documentation? It feels like an anti-pattern to me.

roberth:
Double `mkFlake` has a fundamental problem as discussed earlier. We should not recommend it.
As a secondary concern, this recommendation is incomplete.

PR #156, opened 2023-05-09
Add dogfood template (fix #104 and #121)

### LINKED ISSUE #104: No great way to close over flake inputs

Atry:
## Motivation
- #104
- #121
## Testing
```
nix flake show
```

```
git+file:///Users/twer/github/flake-parts?dir=template/dogfood
├───apps
│   ├───aarch64-darwin
│   │   └───default: app
│   └───x86_64-linux
│       └───default: app
├───flakeModule: unknown
├───flakeModules: unknown
└───packages
    ├───aarch64-darwin
    │   └───hello_22_11: package 'hello-2.12.1'
    └───x86_64-linux
        └───hello_22_11 omitted (use '--all-systems' to show)
```

```
nix repl
```

```
Welcome to Nix 2.15.0. Type :? for help.

nix-repl> :lf .
warning: Git tree '/Users/twer/github/flake-parts' is dirty
Added 20 variables.

nix-repl> flakeModules
{ anotherFlakeModule = { ... }; dogfood = { ... }; hello = { ... }; }

nix-repl> 
```

### DIFF
--- a/template/dogfood/flake.nix
+++ b/template/dogfood/flake.nix
@@ -1,0 +1,17 @@
+{
+  description = "Description for the project";
+
+  inputs = {
+    nixpkgs_22_11.url = "github:NixOS/nixpkgs/nixos-22.11";
+  };
+
+  outputs = { flake-parts, ... }@inputs:
+    flake-parts.lib.mkFlake { inherit inputs; } (
+      flake-parts.lib.mkFlake
+        {
+          # Workaround for https://github.com/hercules-ci/flake-parts/issues/148
+          inputs = inputs // { self.outPath = ./.; };
+        }
+        ./modules/dogfood.nix
+    ).flakeModules.dogfood;
+}
--- a/template/dogfood/flake.nix
+++ b/template/dogfood/flake.nix
@@ -5,12 +5,17 @@
     nixpkgs_22_11.url = "github:NixOS/nixpkgs/nixos-22.11";
   };
 
-  outputs = { flake-parts, ... }@inputs:
+  outputs = { flake-parts, self, ... }@inputs:
     flake-parts.lib.mkFlake { inherit inputs; } (
       flake-parts.lib.mkFlake
         {
+          inherit inputs;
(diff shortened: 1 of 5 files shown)


template/dogfood/flake.nix
```diff
@@ -0,0 +1,22 @@
...
+    flake-parts.lib.mkFlake { inherit inputs; } (
+      flake-parts.lib.mkFlake
+        {
+          inherit inputs;
+          # Workaround for https://github.com/hercules-ci/flake-parts/issues/148
+          self = {
+            outPath = ./.;
+            inherit (self)
+              _type inputs lastModified lastModifiedDate narHash outputs sourceInfo submodules;
+          };
```
roberth, line 18: This breaks normal use of `self`. I don't think importing from `self` is important enough of a problem.
Making assumptions about the user's file layout is too magical. You are free to implement your own conventions using the `specialArgs` solution instead.

roberth (changes requested):
Apart from the design issues, this would need testing and better inline documentation to explain succinctly the role of each file, etc. (Although don't invest in that before making sure everything works with tests, and maybe don't even test before resolving the double mkFlake design issue)
Besides the double mkFlake broken fixpoint, I have doubts about the module keys. A test should point out that that works, e.g. a test doing a double import and a test for disabledModules.

roberth:
Perhaps a way to work around the problem posed by graphs of public flakeModules is to set the module key by hand for now.
I don't think we should bless that solution with a template until we've tried a `mkFlake`-based solution, but it seems that it would unblock your actual module implementation work for the time being.

Issue #157, opened 2023-05-09
`flake.flakeModules` requirements


roberth:
While we have a `flakeModules` option that works decently well in most cases, some requirements aren't satisfied yet, or at least not to a degree where it's easy, let alone obvious how to satisfy them all.

### Easy to dogfood

Due to how the module system `config` fixpoint works, fundamentally, we can't import from the option. This has been [documented](https://flake.parts/dogfood-a-reusable-module.html) but it seems that we'll want to recast this into an easier, considering the intricacies introduced by the other requirements.

### The publishing flake is larger than the dogfooded module

The publishing flake must have the opportunity to define things that are not part of the dogfooded module.
For example, the consumer of the published module generally has no interest in copying the `checks` that come from its dependency. To avoid copying those `checks`, they have to be defined in the publishing flake, but not in the published module.

### Access to the local flake module scope

Modules should be able to access values from the local flake, such as locally packaged build tools with their locked, known-to-work dependencies.

### Deduplication and disabledModules

When references to the local flake must be made, we need to use module syntax that isn't conducive to deduplication and disabledModules.

 - **importApply** does not set a module key, because it is too general for this. It can not assume that its invocations with _different_ arguments should be deduplicated. That would be unexpected; or arbitrary at best.
 - **inline module pattern** does not set a module key. While the module author could write one manually, we'd prefer a foolproof solution. This brings us to...

### Module key for multiple flake modules

 - **flakeModules option** sets a key, but this key is not encountered by modules that import other modules, because the dogfooding requirement (when applicable) does not allow access to the `flakeModules` option.

### Status quo as documented

Does not implement the last requirement, "Module key for multiple flake modules"

###  Double `mkFlake` application

Lets you apply the module key before the main `config` is available, but "access to the local flake module scope" does not work properly.
 
### Finding the local flake by recursing into the `inputs` graph

Leads to scalability issues and spooky action at a distance from deep dependencies; see [Do not traverse inputs](https://flake.parts/best-practices-for-module-writing.html#do-not-traverse-inputs)

Does not seem to solve the dogfooding requirement.

### Special support in `mkFlake`

 - Can do things that a module can't do, e.g. setting top level specialArgs for the purpose of dogfooding.
 - Can use lazy fixpoints instead of badly implemented fixpoint _iteration_-like constructs.
 - Having a more standardized solution seems helpful - even if a solution could be composed from building blocks, it seems that it'd be a tricky composition.
[...]

Issue #148, opened 2023-05-07
Cannot access `inputs.self.outPath`


Atry:
## Given

``` nix
# flake.nix
{
  inputs.flake-parts.url = "github:hercules-ci/flake-parts";
  outputs = inputs@{ self, flake-parts, ... }:
    flake-parts.lib.mkFlake { inherit inputs; } ({inputs, ...}: {
      imports = [
        "${inputs.self.outPath}/my-module.nix"
      ];
    });
}
```

``` nix
# my-module.nix
{
  flake.lib.myFunction = x: x;
}
```

## When running
```
nix flake show
```

## Got

```
error: infinite recursion encountered
```

## Root cause
https://github.com/NixOS/nix/issues/8300

## Workaround

``` nix
# flake.nix
{
  inputs.flake-parts.url = "github:hercules-ci/flake-parts";
  outputs = inputs@{ self, flake-parts, ... }:
    {
      inherit (
        flake-parts.lib.mkFlake { inherit inputs; } ({inputs, ...}: {
          imports = [
            "${inputs.self.outPath}/my-module.nix"
          ];
        })
      ) lib;
    };
}
```

roberth:
## Workaround 2

Instead of `self.outPath`, you could use the equivalent `./.`:

```nix
# flake.nix
{
  inputs.flake-parts.url = "github:hercules-ci/flake-parts";
  outputs = inputs@{ flake-parts, ... }:
    flake-parts.lib.mkFlake { inherit inputs; } ({inputs, ...}: {
      imports = [
        ./my-module.nix
      ];
    });
}
```

Atry:
It's not equivalent when your flake is used in another flake

roberth:
So I guess you're writing a framework with a fixed location for a flake module then?
 
With the Nix we have, you can only `imports` from a `self` where you've already decided the set of outputs attrNames.
In other words, there's no workaround for https://github.com/NixOS/nix/issues/8300.
There might be a solution in the form of https://github.com/NixOS/nix/pull/4154, but otherwise, the best we can do is workarounds.

## Workaround 3, perSystem only

Maybe this helps you or someone. `perSystem` generally does not affect the `attrNames` of the `flake` option.

```nix
# flake.nix
{
  inputs.flake-parts.url = "github:hercules-ci/flake-parts";
  outputs = inputs@{ self, flake-parts, ... }:
    flake-parts.lib.mkFlake { inherit inputs; } ({inputs, ...}: {
[...]

PR #158, opened 2023-05-09
Add moduleLocation to mkFlake argument

Atry:
This PR adds `moduleLocation` argument to mkFlake, allowing the user to set the module key by hand, addressing https://github.com/hercules-ci/flake-parts/pull/156#issuecomment-1540165822

### DIFF
--- a/dev/tests/eval-tests.nix
+++ b/dev/tests/eval-tests.nix
@@ -28,17 +28,17 @@ rec {
 
   emptyExposeArgs = mkFlake
     { inputs.self = { outPath = "the self outpath"; }; }
-    ({ config, moduleLocation, errorLocation, ... }: {
+    ({ config, moduleLocation, ... }: {
       flake = {
-        inherit moduleLocation errorLocation;
+        inherit moduleLocation;
       };
     });
 
   emptyExposeArgsNoSelf = mkFlake
     { inputs.self = throw "self won't be available in case of some errors"; }
-    ({ config, moduleLocation, errorLocation, ... }: {
+    ({ config, moduleLocation, ... }: {
       flake = {
-        inherit moduleLocation errorLocation;
+        inherit moduleLocation;
       };
     });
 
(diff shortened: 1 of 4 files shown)


Atry:
@roberth  Would you mind if I reopen #156 and rebase onto this PR, so that we don't have to break `inputs.self` in order to call `mkFlake` twice.

roberth:
#157 is worth solving, but unless we know that we have a complete solution, we're probably just making a mess out of `mkFlake`.

We already have a "95%" solution of the problem and I am confident that a full solution of #157 is within reach. For this reason, I'm not going to accept a solution that solves a _different_ 95%, and for the sake of maintainability I won't accept a 96% solution either, because that solution will not come with a proof that it leads towards the 100%. If it does come with such a proof, then it's already a 100% solution.

I've already stretched my schedule a lot by responding to your suggestions. I won't be responding quickly for some time, but that won't be too bad, because you can test changes yourself (see e.g. the `dev` directory) and you can validate the requirements or better understand them by applying your solution in multiple repositories. I'm sorry that I won't be quite as helpful for now, but I have to prioritize.

Atry:
This PR does not aim to directly solve #157. Instead, it is a workaround for #148. I understand the best solution of #148 is https://github.com/NixOS/nix/pull/4154. But we probably would not get https://github.com/NixOS/nix/pull/4154 very soon because it is inactive.

I think it would be a net benefit to limit the use of `self.outPath` until https://github.com/NixOS/nix/pull/4154 gets landed, not only because it prevents double `mkFlake`, but also because it would produce obscure error messages about infinite loop due to #148 in other use cases, including `flake-parts`'s own tests.

roberth:
I'd like to keep it simple, so I'm going for the root cause.
- Blocked on https://github.com/NixOS/nix/pull/8908 (or similar)

roberth:
> root cause.
> 
> * Blocked on

Rabbit hole too deep as that issue is itself blocked on another issue.
I think I've found an extended solution that's easier to use.
- #192 

I'm not sure if it (automatically) solves all the problems you intended to solve, but it should solve
- #185

Atry:
Looks good to me

PR #162, opened 2023-05-29
switch to nixpkgs.lib

figsoda:


Kranzes:
What's the benefit?

figsoda:
using `?dir=lib` requires downloading the entire nixpkgs tarball

Kranzes:
I see, will lazy-trees fix this?

figsoda:
I doubt, but don't take my word for it

roberth:
> I see, will lazy-trees fix this?

It can and I [think it should](https://github.com/NixOS/nix/pull/6530#issuecomment-1382268740), but it's actually an orthogonal issue that's only blocked on lazy trees for development reasons. Taking away the parent directory does change the flake semantics, but I think that's for the better. To be continued.

This is going to take a while so I'm open to changing the input until git/github tree fetching is widely available, so I'm warming up to the idea of doing this, having rejected it before.
 - https://github.com/hercules-ci/flake-parts/pull/90

I'll add a comment that explains things.

roberth:
`nixpkgs.lib` contains an outdated version of lib somehow. Maybe its updater is broken.

figsoda:
seems like the actions has been failing, I opened https://github.com/nix-community/nixpkgs.lib/pull/4 to try to fix that

roberth:
I've thought about it for a bit and I really don't want a fragile Action to be in the critical path between flake-parts and module system updates.
You are free to push this in manually with `follows`, but it's not something I wish to push as the default.
Something pinned like an Hercules CI Effect would be an improvement, but I'd rather put my effort into solving the root cause, which is git _tree_ support in `fetchTree` (ironic, considering the naming).

Kranzes:
If you are using flake-parts then you probably are also using nixpkgs, just use `inputs.flake-parts = { url = "github:hercules-ci/flake-parts"; inputs.nixpkgs-lib.follows = "nixpkgs"; }; and it will reuse your nixpkgs
`

figsoda:
yep that's already what I'm doing for my projects, I just noticed it trying to download the whole nixpkgs while debugging #160

bb010g:
Using `nixpkgs.lib` also breaks our `//lib.nix` version check, at least until <https://github.com/nix-community/nixpkgs.lib/pull/5> is merged in our minimum supported `nixpkgs.lib` version.

PR #163, opened 2023-05-29
remove unused code

figsoda:
with the help of deadnix & nil

### DIFF
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -1,4 +1,4 @@
-{ config, lib, flake-parts-lib, ... }:
+{ lib, flake-parts-lib, ... }:
 let
   inherit (lib)
     mkOption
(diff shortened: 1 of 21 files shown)


modules/apps.nix
```diff
@@ -17,11 +17,6 @@ let
   getExe = x:
```
Kranzes, line 17: lib.getExe is now in nixpkgs.

figsoda: the implementation is different, even though I do find it to be better, I want to leave it out of this PR since it is not strictly unused code

Kranzes: Right.

roberth: This `getExe` implementation is a bit picky. That might be a good thing, although it seems to challenge the design goal for flake-parts to be "unopinionated". The underlying goal is to make flake-parts suitable as a foundation for anyone anywhere.
That said, the _existence_ of this logic already makes the options more permissive than the flake schema, but I'd have to check the nix implementation to be sure.

roberth:
While I think there's something to be said for leaving some module arguments in place, I'm happy to accept this cleanup.
Thank you!

bors r+

roberth:
ofcourse

Issue #170, opened 2023-06-14
Would an example folder be welcome?


woile:
Hi everyone, thanks for this awesome project. It does feel more intuitive and simpler to use than flake-utils, and I'm quite happy with it. But I'm struggling with the lack of examples. It took me a while to find how to use `devShells`, and I haven't seen much examples around (I'm relatively new to nix).

My goal would be to see at least one example for each different output supported by `flake-parts`. I could start sending a PR for the `devShells` example, and I was thinking it would be nice if we have examples for:

- running commands (for example a set of commands: `build`, `deploy`)
- configuring formatter (is it something to run things like `black` or `rust fmt`? I cannot tell from the docs)
- apps (what are these for?)
- checks (what are these for?)
- nixosModules (I see one example for `nixosConfigurations`)

roberth:
Hi @woile,

Yes, examples (and/or documentation in general) are very welcome!

I think a good start would be just to do a editing pass over the option docs, but if you prefer to create complete examples, that's also great.

Likely we can improve the infrastructure around examples, templates and tests; so contributions in that area are also welcomed, but not a prerequisite for contributing examples.

> * configuring formatter (is it something to run things like `black` or `rust fmt`? I cannot tell from the docs)

According to upstream docs, it's meant for formatting .nix files, but that may be unenforced.

> I hope the above is correct

Those are definitely some of the outputs! I think formatter is a bit of a weird one, and for the other ones, the existing [option docs](https://flake.parts/options/flake-parts.html) could be a starting point.

There's probably more to discuss, like possible tooling improvements or what to do about other projects' examples, but I *think* I've answered your questions so far and I don't want to bore you ;)

> I could start sending a PR for the `devShells` example

I think that's a great start. Not sure if we need to discuss a lot so it's good to have a starting point and work from there.

srid:
I'd also suggest dedicating a page or section linking to real-world examples. Here's one:

- https://github.com/nix-community/nixd/blob/main/flake.nix

woile:
I created a first PR, let me know what you think 👍🏻 

I like the examples folder because they are great isolated-working examples that just work. In my experience when using rust, the examples folder have been tremendously helpful, and I think it would really help nix get a wider adoption.

We could also use this issue to collect snippets if you want

PR #171, opened 2023-06-15
Add more docs

woile:
This PR adds the following:

- an examples folder
- an example for devShell
- templates are no longer marked as examples, now there are templates and examples avoiding confusion

#170

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,12 @@
 
 _Core of a distributed framework for writing Nix Flakes._
 
-`flake-parts` provides the options that represent standard flake attributes 
+`flake-parts` provides the options that represent standard flake attributes
 and establishes a way of working with `system`.
 Opinionated features are provided by an ecosystem of modules that you can import.
 
 `flake-parts` _itself_ has the goal to be a minimal mirror of the Nix flake schema.
-Used by itself, it is very lightweight. 
+Used by itself, it is very lightweight.
 
 # Why Modules?
 
@@ -71,9 +71,21 @@ then slide `mkFlake` between your outputs function head and body,
 
 Now you can add the remaining module attributes like in the [the template](./template/default/flake.nix).
 
-# Example
+# Templates
 
 See [the template](./template/default/flake.nix).
+
+# Examples
+
+See the folder [examples/](./examples).
+
+# Projects using flake-parts
+
+- [nixd](https://github.com/nix-community/nixd/blob/main/flake.nix) (c++)
+- [hyperswitch](https://github.com/juspay/hyperswitch/blob/main/flake.nix) (rust)
+- [argo-workflows](https://github.com/argoproj/argo-workflows/blob/master/dev/nix/flake.nix) (go)
+- [nlp-service](https://github.com/recap-utr/nlp-service/blob/main/flake.nix) (python)
+- [emanote](https://github.com/srid/emanote/blob/master/flake.nix) (haskell)
 
 # Options Reference
 
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ See [the template](./template/default/flake.nix).
 
 # Examples
 
-See the folder [examples/](./examples).
+See the [examples/](./examples) directory.
 
 # Projects using flake-parts
 
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@ Opinionated features are provided by an ecosystem of modules that you can import
 `flake-parts` _itself_ has the goal to be a minimal mirror of the Nix flake schema.
 Used by itself, it is very lightweight.
 
+---
+
+**Documentation**: [flake.parts](https://flake.parts)
+
+---
+
 # Why Modules?
 
 Flakes are configuration. The module system lets you refactor configuration
@@ -90,7 +96,3 @@ See the folder [examples/](./examples).
 # Options Reference
 
 See [flake.parts options](https://flake.parts/options/flake-parts.html)
-
-# Documentation
-
-See [flake.parts](https://flake.parts)
--- a/examples/project-commands/README.md
+++ b/examples/project-commands/README.md
@@ -1,0 +1,42 @@
+# project-commands
+
+> **Warning**
+> If you copy the flake.nix remember to `git add [-N|--intent-to-add] flake.nix`, otherwise it won't work
+
+This example shows how to create scripts for your project, by leveraging [mission-control](https://github.com/Platonic-Systems/mission-control)
+
+This is a **potential** alternative to:
+
+- Using a `Makefile` to manage your project's scripts
+- Using the popular [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all) approach (having a `scripts/` folder)
+- Using a `bin/` folder
+
+## Explanation
+
+In this example we use the [avro-tools](https://avro.apache.org/) to convert our scripts from `.avdl` to `.avsc`.
+
+You don't need to know anything about avro to understand mission-control and use this example (that's nix baby 🚀).
+
+When setting up [mission-control](https://github.com/Platonic-Systems/mission-control), we add
+one script called `build`. Because of `wrapperName = "run";`, once we open the shell created by nix,
+the commands will be listed as `run build`.
+
+mission-control depends on flake-root, which also exposes the helpful `$FLAKE_ROOT` variable.
+
+After creating the scripts, we need to pass the newly created scripts to the desired shell, in this example we use the default shell.
+
+## Usage
+
+Run:
+
+```sh
+nix develop
+```
+
+And mission-control will print in the new shell the available commands (you should see only one).
+
+Try running
+
+```sh
+run build
+```
--- a/examples/project-commands/README.md
+++ b/examples/project-commands/README.md
@@ -8,14 +8,14 @@ This example shows how to create scripts for your project, by leveraging [missio
 This is a **potential** alternative to:
 
 - Using a `Makefile` to manage your project's scripts
-- Using the popular [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all) approach (having a `scripts/` folder)
-- Using a `bin/` folder
+- Using the popular [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all); a naming convention for a `scripts/` directory
+- Using a `bin/` directory
 
 ## Explanation
 
 In this example we use the [avro-tools](https://avro.apache.org/) to convert our scripts from `.avdl` to `.avsc`.
 
-You don't need to know anything about avro to understand mission-control and use this example (that's nix baby 🚀).
+You don't need to know anything about avro to understand mission-control and use this example (that's Nix baby 🚀).
 
 When setting up [mission-control](https://github.com/Platonic-Systems/mission-control), we add
 one script called `build`. Because of `wrapperName = "run";`, once we open the shell created by nix,
--- a/examples/project-commands/flake.nix
+++ b/examples/project-commands/flake.nix
@@ -1,0 +1,37 @@
+{
+  description = "Description for the project";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+
+    mission-control.url = "github:Platonic-Systems/mission-control";
+    flake-root.url = "github:srid/flake-root";
+  };
+
+  outputs = inputs@{ flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      imports = [
+        inputs.mission-control.flakeModule
+        inputs.flake-root.flakeModule
+      ];
+      systems = [ "x86_64-linux" "aarch64-darwin" "x86_64-darwin" ];
+      perSystem = { config, self', inputs', pkgs, system, ... }: {
+        devShells.default = pkgs.mkShell {
+          nativeBuildInputs = with pkgs; [ avro-tools ];
+          inputsFrom = [ config.mission-control.devShell config.flake-root.devShell ];
+        };
+        mission-control = {
+          wrapperName = "run";
+          scripts = {
+            build = {
+              description = "convert files from .avdl to .avsc";
+              exec = ''
+                avro-tools idl2schemata "$FLAKE_ROOT/Hello.avdl" .
+              '';
+              category = "Development";
+            };
+          };
+        };
+      };
+    };
+}
--- a/examples/shell-environments/README.md
+++ b/examples/shell-environments/README.md
@@ -6,7 +6,7 @@
 This example shows how to create a shell environment which
 includes a diverse set of tools:
 
-```
+```sh
 terraform
 wget
 bat
@@ -29,3 +29,24 @@ You can have as many shells as you want, in this [flake.nix](./flake.nix), you a
 ```sh
 nix develop .#another_env
 ```
+
+## Troubleshooting
+
+### My shell has changed
+
+There 2 possible solutions:
+
+First, using [direnv](https://direnv.net/) to manage your dev environments. See [direnv-guide](https://haskell.flake.page/direnv). This is the recommended approach.
+
+Second is a simple-unreliable hack, which is adding a `shellHook` to `devShells`
+
+```nix
+devShells.default = pkgs.mkShell {
+    shellHook = ''
+    exec $SHELL
+    '';
+};
+```
+
+You might get a lot different issues, use it at your own risk.
+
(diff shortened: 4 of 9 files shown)


README.md
```diff
@@ -71,10 +71,21 @@ then slide `mkFlake` between your outputs function head and body,
...
+
+See the folder [examples/](./examples).
+
+# Projects using flake-parts
+
+- [nixd](https://github.com/nix-community/nixd/blob/main/flake.nix) (c++)
+- [hyperswitch](https://github.com/juspay/hyperswitch/blob/main/flake.nix) (rust)
+- [argo-workflows](https://github.com/argoproj/argo-workflows/blob/master/dev/nix/flake.nix) (go)
+- [nlp-service](https://github.com/recap-utr/nlp-service/blob/main/flake.nix) (python)
+
```
roberth, line 88: I'd add hercules-ci-agent, but that one's too complicated for an example.

```suggestion
- [haskell-flake examples](https://haskell.flake.page/examples) (haskell)
```

@srid Do you have a single, particularly nice to read real world example to put here instead?

examples/shell-environments/README.md
```diff
@@ -0,0 +1,28 @@
...
+wget
+bat
+git
+```
+
+You can search for more package in [nix packages](https://search.nixos.org/packages)
+
+## Usage
+
+The `devShells` maps to the following command:
```
roberth, line 17:
```suggestion
The [`devShells` option](https://flake.parts/options/flake-parts.html#opt-perSystem.devShells) is used by the following command:
```

examples/shell-environments/flake.nix
```diff
@@ -0,0 +1,27 @@
...
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs = inputs@{ flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [ "x86_64-linux" "aarch64-darwin" "x86_64-darwin" ];
+      perSystem = { config, self', inputs', pkgs, system, ... }: {
+        devShells.default = pkgs.mkShell {
+          buildInputs = with pkgs; [ terraform wget bat git ];
```
roberth, line 13:
```suggestion
          nativeBuildInputs = with pkgs; [ terraform wget bat git ];
```

This makes shell completions work.

Also, tl;dr needed for cross and strictDeps.

`buildInputs` is for what I would call "components" or things that can become part of the output. Inputs specified here won't be runnable when enabling `strictDeps` or cross compilation, because the cross compilation infrastructure will pick the variation that has the same build and host platform as the current derivation, which implies that these inputs have a `hostPlatform` that's not compatible with the `buildPlatform`, which is the platform we need for commands we execute.

Seems to complicated to explain here, but I typed it anyway so here you go.

examples/shell-environments/flake.nix
```diff
@@ -0,0 +1,27 @@
...
+        };
+
+        devShells.another_env = pkgs.mkShell {
+          buildInputs = with pkgs; [ curl ];
```
roberth, line 17:
```suggestion
          nativeBuildInputs = with pkgs; [ curl ];
```

examples/shell-environments/flake.nix
```diff
@@ -0,0 +1,27 @@
...
+
+          # we use SHELL to keep the host's shell
+          shellHook = ''
+            $SHELL
+          '';
```
roberth, line 22: This is a neat trick, but I wouldn't recommend it without knowing more about it. Has this been documented somewhere?

```suggestion
```

If you do use this somewhere, I would recommend at least `exec $SHELL` so you don't get returned into a bash and/or the bash shell doesn't stick around. Maybe you need to pass arguments in order to support some `nix develop` features?

I'm not a big fan of hacks like this because `nix develop` itself is already a huge hack. It (and `nix-shell`) try to cover too many use cases, which have incompatible requirements, leading to unsolvable problems and/or a massive amount of options and corner cases. Other shells are _probably_ better covered by something involving `nix print-dev-env`, but that still relies on shell derivations and bash, whereas derivations are not meant to describe shells. I've proposed alternatives, but the implementation work is yet to be done. (and is not something I can allocate for)

modules/devShells.nix
```diff
@@ -17,6 +18,13 @@ mkTransposedPerSystemModule {
       An attribute set of packages to be used as shells.
       [`nix develop .#<name>`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html) will run `devShells.<name>`.
     '';
+    example = literalExpression ''
+    {
+      devShells.default = pkgs.mkShell {
+          buildInputs = with pkgs; [ terraform wget bat git ];
+        };
+    }
```
roberth, line 26: The example should be a valid value for the option, whereas this includes the option path leading to the definition.

```suggestion
    # devShells =
    {
      default = pkgs.mkShell {
        nativeBuildInputs = with pkgs; [ terraform wget bat git ];
      };
    }
```

roberth (commented):
Made a couple suggestions, but overall a good example.
I'll try to be quicker next time.

README.md
srid: If you want real-world example, then: https://github.com/srid/emanote/blob/master/flake.nix

examples/shell-environments/flake.nix
srid: Offtopic, but - to use the host's shell, I'd highly recommend going the direnv route instead, especially with starship (you get a nice visual indication): https://haskell.flake.page/direnv

roberth: That's a great suggestion, and `nix print-dev-env`-based, or based on the same principle, certainly because of how direnv works. Indeed I don't think we should mention that here. It's more something for the site, where we could have a guide for creating a more advanced shell setup.

woile: I will remove it. I agree with your points. I do use starship and using `SHELL` helps me keep it. Otherwise it "feels" strange to me, because as a user I expect to have some packages available, not my shell to change (but I don't have much idea what's happening internally haha). I'm gonna try the `exec $SHELL` because having to exit twice the shell it's annoying haha I've been wondering how to do it.

Going back to the discussion, I'll remove the `$SHELL` and I'm gonna dig a bit into the direnv example. Thanks for the link!

examples/shell-environments/flake.nix
woile: Thanks, it's clera. I was not aware of it! I'll update it 💪🏻

woile:
Thanks for all the awesome feedback, I've updated the PR!

woile:
Added another very easy example with mission-control.

examples/shell-environments/README.md
```diff
@@ -14,6 +14,9 @@ You can search for more package in [nix packages](https://search.nixos.org/packa
 
 ## Usage
 
+> **Warning**
+> If you copy the flake.nix remember to add it to git, otherwise it won't work
```
shanesveller, line 18: Good opportunity to make a proactive suggestion of:
```shell
git add [-N|--intent-to-add] flake.nix
```

examples/project-commands/README.md
```diff
@@ -0,0 +1,42 @@
...
+the commands will be listed as `run build`.
+
+mission-control depends on flake-root, which also exposes the helpful `$FLAKE_ROOT` variable.
+
+After creating the scripts, we need to pass the newly created scripts to the desired shell, in this example we use the default shell.
+
+## Usage
+
+> **Warning**
+> If you copy the flake.nix remember to add it to git, otherwise it won't work
```
shanesveller, line 28: Same suggestion here.

examples/project-commands/flake.nix
```diff
@@ -0,0 +1,37 @@
...
+          nativeBuildInputs = with pkgs; [ avro-tools ];
+          inputsFrom = [ config.mission-control.devShell config.flake-root.devShell ];
+        };
+        mission-control = {
+          wrapperName = "run";
+          scripts = {
+            build = {
+              description = "convert files from .avdl to .avsc";
+              exec = ''
+                avro-tools idl2schemata "$FLAKE_ROOT/Hello.avdl" .
```
srid, line 29: `$FLAKE_ROOT/` here is redundant, since mission-control always runs commands from project root (unless `cdToProjectRoot` is turned off).

woile: I've been trying to remove it. But the moment I remove the script stops working.

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.io.File.toURI()" because the return value of "java.io.File.getParentFile()" is null
        at org.apache.avro.compiler.idl.Idl.<init>(Idl.java:53)
        at org.apache.avro.tool.IdlToSchemataTool.run(IdlToSchemataTool.java:47)
        at org.apache.avro.tool.Main.run(Main.java:67)
        at org.apache.avro.tool.Main.main(Main.java:56)
```

examples/shell-environments/README.md
```diff
@@ -0,0 +1,52 @@
...
+> If you copy the flake.nix remember to `git add [-N|--intent-to-add] flake.nix`, otherwise it won't work
+
+This example shows how to create a shell environment which
+includes a diverse set of tools:
+
+```sh
+terraform
+wget
+bat
+git
```
roberth, line 13: I don't think we should include `git`, but rather rely on the user's git.
They need to have their own git in their user profile in order to clone new projects, so it's not necessary to provide one here, and having potentially multiple versions of git operating on the same repos is a *bit* asking for trouble, because you're essentially doing a whole bunch of upgrades and downgrades all the time, if you want to look at it that way. Git is of course managed very well, so you might get away with it, but git in a shell could prove problematic later.

How about `nixpkgs-fmt` instead?

examples/project-commands/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+- Using a `Makefile` to manage your project's scripts
+- Using the popular [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all) approach (having a `scripts/` folder)
+- Using a `bin/` folder
+
+## Explanation
+
+In this example we use the [avro-tools](https://avro.apache.org/) to convert our scripts from `.avdl` to `.avsc`.
+
+You don't need to know anything about avro to understand mission-control and use this example (that's nix baby 🚀).
```
roberth, line 18:
```suggestion
You don't need to know anything about avro to understand mission-control and use this example (that's Nix baby 🚀).
```

examples/project-commands/flake.nix
roberth: Should be solved by Nix, fyi
- https://github.com/NixOS/nix/issues/8034

examples/shell-environments/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+You can search for more package in [nix packages](https://search.nixos.org/packages)
```
roberth, line 16:
```suggestion
You can search for more packages in [nix packages](https://search.nixos.org/packages)
```

examples/shell-environments/README.md
```diff
@@ -0,0 +1,52 @@
...
+You can have as many shells as you want, in this [flake.nix](./flake.nix), you also have
+`another_env` which includes `curl`. To open it:
+
+```sh
+nix develop .#another_env
+```
+
+## Troubleshooting
+
+### My shell has changed
```
roberth, line 35:
```suggestion
### I get bash instead of my shell
```

examples/shell-environments/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+There 2 possible solutions:
```
roberth, line 37:
```suggestion
`nix develop` was designed for Nixpkgs stdenv, which uses bash, so that you can troubleshoot a Nix build with it. If you use a different shell, you'll want to get just the variables instead.

There are 3 possible solutions:
```

examples/shell-environments/README.md
```diff
@@ -0,0 +1,52 @@
...
+```nix
+devShells.default = pkgs.mkShell {
+    shellHook = ''
+    exec $SHELL
+    '';
+};
+```
+
+You might get a lot different issues, use it at your own risk.
+
```
roberth, line 52:
```suggestion

Lastly, there's `nix print-dev-env` which returns the variables - in case you're feeling adventurous, because this is far from a complete solution. See `nix print-dev-env --help`.
```

examples/shell-environments/flake.nix
```diff
@@ -0,0 +1,22 @@
...
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs = inputs@{ flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [ "x86_64-linux" "aarch64-darwin" "x86_64-darwin" ];
+      perSystem = { config, self', inputs', pkgs, system, ... }: {
+        devShells.default = pkgs.mkShell {
+          nativeBuildInputs = with pkgs; [ terraform wget bat git ];
```
roberth, line 13:
```suggestion
          nativeBuildInputs = with pkgs; [ terraform wget bat nixpkgs-fmt ];
```

To match the README suggestion earlier.

examples/project-commands/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+> **Warning**
+> If you copy the flake.nix remember to `git add [-N|--intent-to-add] flake.nix`, otherwise it won't work
+
+This example shows how to create scripts for your project, by leveraging [mission-control](https://github.com/Platonic-Systems/mission-control)
+
+This is a **potential** alternative to:
+
+- Using a `Makefile` to manage your project's scripts
+- Using the popular [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all) approach (having a `scripts/` folder)
```
roberth, line 11:
```suggestion
- Using the popular [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all); a naming convention for a `scripts/` directory
```

A bit more info, and folder -> directory, which is the term we generally use in and around Nix.

README.md
```diff
@@ -71,14 +77,22 @@ then slide `mkFlake` between your outputs function head and body,
...
-# Example
+# Templates
 
 See [the template](./template/default/flake.nix).
 
-# Options Reference
+# Examples
 
-See [flake.parts options](https://flake.parts/options/flake-parts.html)
+See the folder [examples/](./examples).
```
roberth, line 86:
```suggestion
See the [examples/](./examples) directory.
```

Nix uses "directory"

examples/project-commands/README.md
```diff
@@ -0,0 +1,42 @@
...
+- Using a `bin/` folder
```
roberth, line 12:
```suggestion
- Using a `bin/` directory
```

modules/devShells.nix
```diff
@@ -17,6 +18,13 @@ mkTransposedPerSystemModule {
       An attribute set of packages to be used as shells.
       [`nix develop .#<name>`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html) will run `devShells.<name>`.
     '';
+    example = literalExpression ''
+    {
+      devShells.default = pkgs.mkShell {
+          nativeBuildInputs = with pkgs; [ wget bat git cargo ];
+        };
+    }
```
roberth, line 26:
```suggestion
    # devShells =
    {
      default = pkgs.mkShell {
        nativeBuildInputs = with pkgs; [ wget bat cargo ];
      };
    }
```

Module docs don't repeat the path to the option.
No git.
Consistent indentation.

examples/shell-environments/README.md
```diff
@@ -0,0 +1,52 @@
...
+First, using [direnv](https://direnv.net/) to manage your dev environments. See [direnv-guide](https://haskell.flake.page/direnv). This is the recommended approach.
+
+Second is a simple-unreliable hack, which is adding a `shellHook` to `devShells`
+
+```nix
+devShells.default = pkgs.mkShell {
+    shellHook = ''
+    exec $SHELL
+    '';
```
roberth, line 47:
```suggestion
  shellHook = ''
    exec $SHELL
  '';
```

Nix expressions are usually indented with 2 spaces.
You can indent your `''` string contents and have an equivalent expression.

examples/project-commands/flake.nix
woile: What should we do then? Shall we leave it for now? I'm gonna follow the issue as well

roberth: I think it's fine to leave as is for now.

woile:
Everything updated 🚀

roberth (approved):

roberth:
Thank you @woile!

PR #180, opened 2023-07-31
apps: use lib.getExe to find executable

patricksjackson:
Use nixpkgs version of [getExe](https://github.com/NixOS/nixpkgs/blob/39ae486d8caecdf1a057ae334381e477b93ebde9/lib/meta.nix#L146) to determine the executable name.

This brings it inline with the behavior of [the rest of the ecosystem](https://github.com/NixOS/nix/blob/c0e735f474a3a581cb08e8aa54e0001c9809ee92/src/nix/run.md?plain=1#L49-L51), which allows a fallback to the pname/name.

I tested this with a local repo to confirm.

### DIFF
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -3,6 +3,7 @@ let
   inherit (lib)
     mkOption
     types
+    getExe
     ;
   inherit (flake-parts-lib)
     mkTransposedPerSystemModule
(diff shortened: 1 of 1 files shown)


roberth:
It's convenient but can lead to bad error messages; at least as currently implemented.
I've started a discussion at https://github.com/NixOS/nixpkgs/pull/167247/files#r1279561455 and at least the author is positive about it.

patricksjackson:
No worries, I did notice the upstream discussion. I was under the impression that `meta.mainProgram` wasn't set for most of nixpkgs, so I'm happy to see that it was fairly easy to add it to the few missing places.

I'm not sure what your plan is here, but I think the best option would probably be to match [your new version in nixpkgs](https://github.com/NixOS/nixpkgs/pull/246386), so we'd warn but still allow the fallback at least for now. Whether that's done by using this current change or by rewriting that new impl here for backwards compat with older nixpkgs would be up to you.

Feel free to take over this and do whatever. I don't have major objections to any of the options (including just closing this PR) now that it's clear `meta.mainProgram` is the agreed upon standard/required way of declaring the name.

PR #183, opened 2023-08-18
lib/`mkFlake`: Document `_file` infinite recursion

bb010g:
Split out from #137. Please read the individual commit messages.


roberth:
You mean https://github.com/NixOS/nix/pull/7796/commits/5d834c40d0a1e397cc650f88b1544ee2e5912400

> both `outPath` and `sourceInfo` would have to be inputs

I don't think we *need* `sourceInfo`, but perhaps its metadata would be nice to use. We should be careful with metadata though, because differentiating by origin can be a double edged sword. We have to make sure that inputs are interchangeable: a fork should work just as well as the original. (Not saying this is an immediate problem; just raising awareness of the requirement)

> would have to be inputs to `outputs` that don't force `self` in order to consistently provide proper error messages when evaluation of `self` errors?

Sounds about right. I think `outPath` would be a very unfortunate name for such a parameter.
A similar effect could be achieved by asking the caller to say something like `mkFlake { here = ./.; ... }`.

Not sure if we should act on the "here" idea now. For most users it's probably better to get a change into Nix so we can improve this structurally without bothering them with an extra parameter.

dev/tests/eval-tests.nix
```diff
@@ -2,59 +2,257 @@
...
+        flake.sourceInfo or { };
+      inputs = flake.inputs or { };
+      outputs = flake.outputs (inputs // { self = result; });
+      result = outputs // sourceInfo // {
+        inherit inputs outputs sourceInfo;
+        _type = "flake";
+      };
+    in
+    assert builtins.isFunction flake.outputs;
+    result;
```
roberth, line 39: Do we really need `weakEvalTests.callFlake`?

Not happy about how complicated this got. Relying on laziness and overriding means that it's not immediately obvious what runs how.

dev/flake-module.nix
```diff
@@ -38,11 +44,18 @@
...
+    # for REPL exploration / debugging
+    config.allFlakeModuleArgs = args // config._module.args // specialArgs;
+    config.allRootModuleArgs = rootArgs // rootConfig._module.args // rootSpecialArgs;
+    config.transformFlakeModule = flakeModuleTransformer:
+      rootSpecialArgs.replaceSpecialArgs (prevSpecialArgs: prevSpecialArgs // {
+        inherit flakeModuleTransformer;
+      });
+    options.mySystem = lib.mkOption {
+      default = rootConfig.allSystems.${builtins.currentSystem};
+    };
```
roberth, line 57: Can't we just set `config.debug = true;`?

`dev` is already overly complicated because it tries to work around the subflake problems. I'd like to simplify it instead of obfuscating it further.

bb010g:
> I don't think we _need_ `sourceInfo`, but perhaps its metadata would be nice to use. We should be careful with metadata though, because differentiating by origin can be a double edged sword. We have to make sure that inputs are interchangeable: a fork should work just as well as the original. (Not saying this is an immediate problem; just raising awareness of the requirement)

Alright. Balancing this is hard when debug information is sourced the same as normal information.

> Not sure if we should act on the "here" idea now. For most users it's probably better to get a change into Nix so we can improve this structurally without bothering them with an extra parameter.

I agree on this. A manual parameter that they have to set to `./flake.nix` is definitely not what this should be long-term.

dev/tests/eval-tests.nix
```diff
@@ -2,59 +2,257 @@
...
+  # `self` will never evaluate *and* `builtins.toString self.outPath` must
+  # evaluate, causing Nix to instead throw an infinite recursion error. Even
+  # just `inputs.self ? outPath` throws an infinite recursion error.
+  # (`builtins.tryEval` can only catch errors created by `builtins.throw` or
+  # `builtins.assert`, so evaluation is guarded with
+  # `exhibitingInfiniteRecursion` here to keep `runTests` from diverging.)
+  # In this particular case, `mkFlake` evaluates `self ? outPath` to know if the
+  # default module location it provides should be generic or specific. As
+  # explained, this evaluation is unsafe under an uncatchably divergent `self`.
+  # Thus, `outPath` cannot be safely sourced from `self` at the top-level.
```
roberth, line 195: :tada: I can confirm that this now works as it should, since #192 or #194.

<details><summary>The new, informative error message</summary>

```
«error: error:
       … while evaluating the attribute 'strongEvalTests.nonexistentOption'

         at /home/user/h/flake-parts/dev/tests/eval-tests.nix:209:3:

          208|   # exhibiting infinite recursion, the flake is made equivalent to `empty`.
          209|   strongEvalTests.nonexistentOption = evalTests.callFlake {
             |   ^
          210|     inputs.flake-parts = evalTests.flake-parts;

       … in the left operand of the update (//) operator

         at /home/user/h/flake-parts/dev/tests/eval-tests.nix:33:24:

           32|       outputs = flake.outputs (inputs // { self = result; });
           33|       result = outputs // sourceInfo // {
             |                        ^
           34|         inherit inputs outputs sourceInfo;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: The option `nonexistentOption' does not exist. Definition values:
       - In `/home/user/h/flake-parts/dev/tests/eval-tests.nix': null»
```

</details>

PR #192, opened 2023-10-13
moduleLocation, but also automatic

roberth:
This improves on #158 by making the default module location automatic.

When `moduleLocation` is specified, it behaves as in #158.
If not, the default module location is determined from the `inputs` attribute position, with fallback.

I believe `moduleLocation` is sufficient for the use cases in the flake-parts modules that were changed.
I am currently not aware of a case where `errorLocation` would have to be exposed to the modules, so I'd like to keep the interface minimal until we find that.


Atry (approved):
Looks good to me

PR #197, opened 2023-11-03
Add an example of dogfooding (fix #121)

### LINKED ISSUE #121: Exposing a flakeModule that depends on other flake modules without defining in the same file

Atry:
This PR is a rework of https://github.com/hercules-ci/flake-parts/pull/156. It avoids infinite loops by using partitions.

### DIFF
--- a/template/dogfood/flake.nix
+++ b/template/dogfood/flake.nix
@@ -2,7 +2,7 @@
   description = "Description for the project";
 
   inputs = {
-    nixpkgs_23_05.url = "github:NixOS/nixpkgs/nixos-23.05";
+    nixpkgs_24_05.url = "github:NixOS/nixpkgs/nixos-24.05";
   };
 
   outputs = { flake-parts, self, nixpkgs, ... }@inputs:
--- a/template/dogfood/modules/custom-hello.nix
+++ b/template/dogfood/modules/custom-hello.nix
@@ -12,19 +12,19 @@
       options.customHello.enableUserStdenv = lib.mkEnableOption "stdenv from `flakeModules.customHello` user's nixpkgs";
       options.perSystem = flake-parts-lib.mkPerSystemOption ({ pkgs, system, ... }: {
         packages.hello =
-          (inputs.nixpkgs_23_05.legacyPackages.${system}.hello.override {
+          (inputs.nixpkgs_24_05.legacyPackages.${system}.hello.override {
             stdenv =
               if cfg.enableUserStdenv then
                 pkgs.stdenv
               else
-                inputs.nixpkgs_23_05.legacyPackages.${system}.stdenv;
+                inputs.nixpkgs_24_05.legacyPackages.${system}.stdenv;
           }).overrideAttrs (oldAttrs: {
             meta = oldAttrs.meta // {
-              description = "A hello package from the `flakeModules.customHello` author's nixpkgs 23.05, built with stdenv from ${
+              description = "A hello package from the `flakeModules.customHello` author's nixpkgs 24.05, built with stdenv from ${
               if cfg.enableUserStdenv then
                 "the `flakeModules.customHello` user's nixpkgs"
               else
-                "the `flakeModules.customHello` author's nixpkgs 23.05"
+                "the `flakeModules.customHello` author's nixpkgs 24.05"
             }";
             };
           });


template/dogfood/flake.nix
```diff
@@ -0,0 +1,20 @@
+{
+  description = "Description for the project";
+
+  inputs = {
+    nixpkgs_22_11.url = "github:NixOS/nixpkgs/nixos-22.11";
```
shanesveller, line 5: 23.11 is targeting release by EOM and 23.05 has been available for quite a while too, shouldn't this be updated to something newer?

Atry: It's just an example of specifying a special version of `nixpkgs`

shanesveller: Just a bystander here so you can skip the feedback if you prefer, but I would expect a `template` output (as opposed to a documentation-style prose example, which can be more relaxed) to produce a well-formed, actionable flake that doesn't include any subjectively-bad defaults.

Atry: fixed

Wily96 (approved):
// FindOptions are for fetching Artifact(s) out of the scope of the current run.
// Must specify a PAT with actions:read scope for cross run/repo lookup otherwise these will be ignored.
export interface FindOptions {
  findBy?: {
    // Token with actions:read permissions
    token: string
    // WorkflowRun of the artifact(s) to lookup
    workflowRunId: number
    // Repository owner
    repositoryOwner: string
    // Repository name
    repositoryName: string
  }
}

template/dogfood/flake.nix
```diff
@@ -0,0 +1,27 @@
...
+      (topLevel: {
+        imports = [
+          flake-parts.flakeModules.partitions
+          ./modules/dev.nix
+        ];
+
+        systems = [ "x86_64-linux" "aarch64-darwin" ];
+
+        partitionedAttrs.devShells = "dogfood";
+        partitionedAttrs.packages = "dogfood";
```
roberth, line 22: Having to list these is not great UX.

I think I'd prefer to add a parameter to `mkFlake`.
I think we could have a `modules` specialArg to make that nice:

```nix
mkFlake {
  inherit inputs;
  publicModules.foo = ./foo.nix;
}
({ modules, ... }: {
  imports = [
    modules.self.foo  # based on the publicModules parameter
    modules.git-hooks-nix.default  # just a mapAttrs of inputs; lazy and cheap
  ];
})
```

It's a bit more custom, but it avoids the complexity of having partitions, probably improving error messages as well, and slightly less boilerplate.

It does put the module in a different scope, but I think this can be worked around, with techniques similar to `withSystem`, and it'd look like:

```nix
  publicModules = local@{ config, withSystem, ... }: { # not defining a module; just a function
  
    /** A module that adds a check using a package built in the local way,
        with no interference except `follows` */
    foo = userFlake@{ self, ... }: {
      perSystem = { system, pkgs, ... }:
        withSystem system (localPerSystem@{ ... }: {
          checks.foo = pkgs.runCommand "check-foo" {
            nativeBuildInputs = [ localPerSystem.config.packages.foo ];
          } "foo ${userFlake.self} && touch $out";
        });
    };
  };
```

However, actually needing access to the local flake seems quite rare for flake-parts modules, so we might not even need it for now.

What do you think?

Atry: I think a flake module that defines a flake module is a good abstract, because it distinguishes the module author's inputs from the module user's inputs. I like the idea to make `publicModules` a partition of preparatory step by default before the "main" partition. This design has precedent in other languages, e.g. [sbt](https://www.scala-sbt.org/1.x/docs/Organizing-Build.html).

Atry: The reason why we cannot put bootstrap modules to `flakeModules` is that `flakeModules` is a submodule of the `mkFlake`'s top level module, and the `mkFlake`'s top level module cannot depend on its own submodule.

Can we make `mkFlake` a submodule instead of a function, then we can create a "meta-module" to set up special args of the `mkFlake`. Eventually the user could simply use nixpkgs's `evalModules` to evaluate the meta-module to build the flake:
```nix
(lib.evalModules {
  modules = [
    inputs.flake-parts.modules.metaModule
    {
      inherit inputs;
      publicModules.myPublicModule = {};
      mkFlake = {modules, flake-parts-lib, lib, inputs, ...}: {
        imports = [ modules.myPublicModule ];
        perSystem = { ... };
      };
    }
  ];
}).config.mkFlake.flake
```

PR #200, opened 2023-12-10
raise error when flake attribute doesn't exist

DavHau:
Currently when trying to access packages from input flakes for a system that isn't supported by that flake, the error is not very conclusive.
It costed me at least an hour to grasp this. Now putting in this error message to prevent others from falling into the same trap.

### DIFF
--- a/modules/transposition.nix
+++ b/modules/transposition.nix
@@ -67,7 +67,11 @@ in
     perInput =
       system: flake:
       mapAttrs
-        (attrName: attrConfig: flake.${attrName}.${system})
+        (attrName: attrConfig:
+          flake.${attrName}.${system} or (throw ''
+            Attemt to access non existent attribute ${attrName}.${system} of flake ${flake}.
+          '')
+        )
         config.transposition;
 
     perSystem = {


modules/transposition.nix
```diff
@@ -67,7 +67,11 @@ in
     perInput =
       system: flake:
       mapAttrs
-        (attrName: attrConfig: flake.${attrName}.${system})
+        (attrName: attrConfig:
+          flake.${attrName}.${system} or (throw ''
+            Attemt to access non existent attribute ${attrName}.${system} of flake ${flake}.
+          '')
```
roberth, line 73: Often this will be recoverable. For instance for `packages`, we can return `{ }`, so that `inputs'.foo.packages?bar` works.

```suggestion
          flake.${attrName}.${system} or (
            if perInputDefault?${attrName}
            then perInputDefault.${attrName} system
            else throw ''
                Attemt to access non-existent attribute ${lib.escapeNixIdentifier attrName}.${lib.escapeNixIdentifier system} of flake input ${lib.escapeNixIdentifier flake}.
              '')
```

roberth (commented):
A good improvement, and have a suggestion.

modules/transposition.nix
```diff
@@ -67,7 +67,11 @@ in
     perInput =
       system: flake:
       mapAttrs
-        (attrName: attrConfig: flake.${attrName}.${system})
+        (attrName: attrConfig:
+          flake.${attrName}.${system} or (throw ''
+            Attemt to access non existent attribute ${attrName}.${system} of flake ${flake}.
```
msgilligan, line 72: "Attempt"

msgilligan (changes requested):
Minor typo should be fixed.

modules/transposition.nix
roberth: Thanks!
I've mostly redone the error message in #206 to add more info and fix the problem with `${flake}`.

PR #210, opened 2024-01-29
templates: make flake-parts input explicit

MatthewCroughan:
Making the input implicit and relying on the registry has never worked for me, and I don't like relying on it

```
nix flake show
warning: Git tree '/home/matthew/tmp/llm-chain' is dirty
error:
       … while updating the lock file of flake 'git+file:///home/matthew/tmp/llm-chain'

       … while updating the flake input 'flake-parts'

       error: cannot find flake 'flake:flake-parts' in the flake registries
```

Each new template would just do this for me, and I always end up having to write the input explicitly anyway.

### DIFF
--- a/dev/tests/template.nix
+++ b/dev/tests/template.nix
@@ -15,18 +15,17 @@ hci-effects.mkEffect {
     nix -v flake init -t ${../..}
 
     ann pointing to local sources...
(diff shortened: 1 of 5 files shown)


roberth (approved):
Explicit is good.

roberth:
Thanks!

PR #213, opened 2024-02-16
Add flake.modules (opt-in)

roberth:
Allow modules of any class to be published.
This removes the need for custom flake attributes, such as `nixosModules`, `flakeModules`, etc.

- Depends on #154 (somewhat)

### DIFF
--- a/dev/flake-module.nix
+++ b/dev/flake-module.nix
@@ -1,4 +1,4 @@
-{ config, lib, inputs, withSystem, ... }:
+{ config, lib, inputs, self, withSystem, ... }:
 
 {
   imports = [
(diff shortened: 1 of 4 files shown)


cdmistman (approved):
just tested and confirmed it works for my use case! thank you!

Atry:
Any issue preventing this PR from being merged?

roberth:
[Documentation at flake.parts](https://flake.parts/options/flake-parts-modules)

srid:
This would be considered deprecated right?

https://flake.parts/options/flake-parts-flakemodules

roberth:
I believe `nix flake check` does not accept `modules` yet, so I wouldn't recommend to migrate yet, but feel free to start using `modules`.

PR #221, opened 2024-04-14
docs: recommend setting `moduleLocation` by default

Atry:
As `moduleLocation` can be used to avoid portential infinite loops issue, it would be nice to mention it from README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,10 @@ then slide `mkFlake` between your outputs function head and body,
 
 ```nix
   outputs = inputs@{ flake-parts, ... }:
-    flake-parts.lib.mkFlake { inherit inputs; } {
+    flake-parts.lib.mkFlake {
+      inherit inputs;
+      moduleLocation = ./flake.nix;
+    } {
       flake = {
         # Put your original flake attributes here.
       };


roberth:
Is that still needed? Did you run into this again?
If users set `inputs` (which almost certainly they do), Nix will use the attribute position to determine the file.
That's since #192; half a year ago.

Atry:
`moduleLocation` can be used to locate `./.` to avoid stack overflow. See https://github.com/Atry/nix-ml-ops/blob/f37e70da4b6a26dbf8ca06741b87b2676f6996d5/flake-modules/poetry2nix.nix#L55-L59

Atry:
Currently flake modules do not have the ability to visit `./.`. `self.outPath` is similar but it can result in infinite loops and stack overflow.

roberth:
> Currently flake modules are not able to visit `./.`. `self.outPath` is similar but it can result in infinite loops and stack overflow.

It seems to work fine nowadays. Example: https://github.com/srid/haskell-flake/blob/e47512f9a1eda70c762a0265cf029ecdaa9ffdf3/nix/modules/project/default.nix#L19-L29

Do you have an example where `moduleLocation` must be set?

I'd like to avoid a situation where we make unnecessary recommendations, and I'm making an effort to make Nix provide the info robustly without `unsafeGetAttrPos` or `self`. (https://github.com/NixOS/nix/pull/8908, currently blocked on a decision for https://github.com/NixOS/nix/issues/10350)

PR #223, opened 2024-04-26
add darwinModules

cdmistman:
this is very similar to the nixosModules module, only with some modified copy. One thing i'll note is that the type is `lazyAttrsOf unspecified` when it might be better as `lazyAttrsOf submodule`. Regardless, this would be good to make darwin modules more accessible like nixos modules :)


cdmistman:
closing after discussion in #224

PR #235, opened 2024-07-03
Test mkFlake specialArgs

roberth:
It already worked, but didn't have a test in `eval-tests.nix`.

### DIFF
--- a/dev/tests/eval-tests.nix
+++ b/dev/tests/eval-tests.nix
@@ -135,6 +135,16 @@ rec {
     config = { };
   };
 
+  specialArgFlake = mkFlake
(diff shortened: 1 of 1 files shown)


srid:
Interesting. Is there a real-world use case for `specialArgs` in `mkFlake`? Because it runs outside of perSystem context, I wonder what sort of things the user would want to pass that are not `inputs`.

roberth:
It could be used to override something, or maybe as part of some integration of flake-parts into something else.
I don't see a particularly good use case for it, but it's been a public interface for a long time now, so it's better to make sure it keep working.

PR #240, opened 2024-08-07
apps: Add `meta` option

shivaraj-bh:
This allows specifying metadata info about the flake app. This has been standardised in Nix at https://github.com/NixOS/nix/pull/11297

Here’s an example of its usage:
```nix
# Inside perSystem
{
  apps.hello = {
    program = pkgs.hello;
    type = "app";
    meta.description = "I say hello!";
  };
}
```

### DIFF
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -34,6 +34,16 @@ let
           A path to an executable or a derivation with `meta.mainProgram`.
         '';
       };
+      meta = mkOption {
+        type = types.lazyAttrsOf lib.types.raw;
+        default = { };
+        description = ''
+          Metadata information about the app.
+          Standardaized in Nix at <https://github.com/NixOS/nix/pull/11297>.
+
+          Note: `nix flake check` is only aware of the `description` attribute in `meta`.
+        '';
+      };
     };
   };
 in
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -37,6 +37,7 @@ let
       meta = mkOption {
         type = types.lazyAttrsOf lib.types.raw;
         default = { };
+        # TODO refer to Nix manual 2.25
         description = ''
           Metadata information about the app.
           Standardized in Nix at <https://github.com/NixOS/nix/pull/11297>.
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -39,7 +39,7 @@ let
         default = { };
         description = ''
           Metadata information about the app.
-          Standardaized in Nix at <https://github.com/NixOS/nix/pull/11297>.
+          Standardized in Nix at <https://github.com/NixOS/nix/pull/11297>.
 
           Note: `nix flake check` is only aware of the `description` attribute in `meta`.
         '';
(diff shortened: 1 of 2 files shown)


roberth:
I'm supportive of the idea, but the Nix manual [states](https://nix.dev/manual/nix/2.23/command-ref/new-cli/nix3-run.html?highlight=apps#apps)

> The only supported attributes are:
> * `type` [...]
> * `program` [...]

I guess this is somewhat open to interpretation, but allowing any attribute whatsoever does not seem to be the intent.
`nix flake check` is silent about it, but most likely that's just an oversight.

Could you open a PR to nix that documents the `description` attribute? That way it can become a standard, and everyone benefits from better metadata and discoverability.

As for this PR, I'd be happy to accept a `description` option when the Nix PR is accepted, but preferably not a `freeformType`, because that would let typos in these attribute names to go uncaught, and iirc a `freeformType` can also cause worse errors when "recursive" definitions are involved (e.g. `desrcription = "something ${baseNameOf config.command}"` would be an infinite recursion because of the typo+freeformType).

shivaraj-bh:
> Could you open a PR to nix that documents the description attribute? That way it can become a standard, and everyone benefits from better metadata and discoverability.

Good idea! Here’s the PR: https://github.com/NixOS/nix/pull/11297

shivaraj-bh:
@roberth I have made changes based on what was standardised upstream

modules/apps.nix
```diff
@@ -34,6 +34,16 @@ let
           A path to an executable or a derivation with `meta.mainProgram`.
         '';
       };
+      meta = mkOption {
+        type = types.lazyAttrsOf lib.types.raw;
+        default = { };
+        description = ''
+          Metadata information about the app.
+          Standardaized in Nix at <https://github.com/NixOS/nix/pull/11297>.
```
roberth, line 42:
```suggestion
          Standardized in Nix at <https://github.com/NixOS/nix/pull/11297>.
```

modules/apps.nix
```diff
@@ -34,6 +34,16 @@ let
           A path to an executable or a derivation with `meta.mainProgram`.
         '';
       };
+      meta = mkOption {
+        type = types.lazyAttrsOf lib.types.raw;
+        default = { };
+        description = ''
```
roberth, line 40:
```suggestion
        # TODO refer to Nix manual 2.25
        description = ''
```

roberth (commented):
I was thinking ideally `meta` could be a submodule with freeform type, but that would cause `description` to error or have a default like `""`, neither of which is an improvement, so this is probably the best solution for now.

Related:
- optional options could help out with this https://github.com/NixOS/nixpkgs/pull/334680

shivaraj-bh:
@roberth Does it look good now?

srid:
@roberth Is there anything to be done here before merging? This would be useful to implement: https://github.com/juspay/omnix/issues/163

roberth (approved):

modules/apps.nix
```diff
@@ -34,6 +34,17 @@ let
           A path to an executable or a derivation with `meta.mainProgram`.
         '';
       };
+      meta = mkOption {
+        type = types.lazyAttrsOf lib.types.raw;
+        default = { };
```
srid, line 39: In hindsight, should we add a default here? Taking from `program`, that is, if it is a derivation. But `coercedTo` seems lossy, so this probably is not possible.

Until then, users basically will have to explicit set it like follows:

image

PR #242, opened 2024-08-29
Partitions

roberth:
- Adds a mechanism for #119 
- Make `nix develop` work properly
- Get rid of custom hacks to keep dev deps out of the main `flake-parts` lock file


Atry:
`partition` looks like a general mechanism, can we upstream it to `nixpkgs`?

roberth:
What would be another use case for it? If a configuration doesn't have vastly different use cases that require significantly different sets of fetched inputs, it serves little to no purpose and is not worth the complexity.

It is quite general, but looking at the implementation, I don't see much opportunity to generalize it. A lot of it is actually quite specific to flakes, when looking at the names, the extra inputs, the options descriptions.
So perhaps it's better for it to be a pattern rather than an abstraction.

PR #248, opened 2024-09-12
Remove the default value of `extraInputsFlake`

Atry:
`options.extraInputsFlake.isDefined` is always `true` if there is a default value for `extraInputsFlake`, which might not be desired.


roberth:
Indeed that default did not belong there. Thanks!

PR #249, opened 2024-09-12
Set proper type for `flake.nixosModules`

Atry:


roberth (approved):
Thanks!

phip1611:
This changes the structure of `<flake>.nixosModules` - is this intended? Shouldn't this be at least documented in the Changelog?

Before, 9c92fd1582b5b025c5c9ec86878618c662718288, `nixosModules.<name>.imports` evaluated to something like
"/nix/store/87r00nfycvkkhxn94mn1xpwr6vrvc7pj-common/modules/<name>"`

Now, I have to use `builtins.head (builtins.head (builtins.head nixosModules.bootitems.imports).imports).imports`

to get the same output.

roberth:
I accepted this change under the assumption that module would be used with the module system and not dissected with custom functions.

I wouldn't mind adding a function that flattens these wrapper modules into a single layer, or a new function in the module system for the purpose of adding these attributes without wrapping the module, but ideally we'd treat the modules as black boxes and just use existing functions.

Perhaps you could achieve your goal by calling `pkgs.nixosOptionsDoc` or by exposing something from within the NixOS documentation module that calls that function?

phip1611:
I have a (rather obscure?) use-case where I need the file paths: https://github.com/phip1611/nixos-configs/commit/64755a81f6d4a68653b96e1a0de834bd0d637021

PR #251, opened 2024-09-12
Set `key` for `modules` and `nixosModules`

Atry:
Fixes https://github.com/NixOS/nixpkgs/issues/340361

### DIFF
--- a/modules/nixosModules.nix
+++ b/modules/nixosModules.nix
@@ -15,7 +15,11 @@ in
       nixosModules = mkOption {
         type = types.lazyAttrsOf types.deferredModule;
         default = { };
-        apply = mapAttrs (k: v: { _file = "${toString moduleLocation}#nixosModules.${k}"; imports = [ v ]; });
+        apply = mapAttrs (k: v: {
+          _file = "${toString moduleLocation}#nixosModules.${k}";
+          key = "${toString moduleLocation}#nixosModules.${k}";
+          imports = [ v ];
+        });
         description = ''
           NixOS modules.
 
--- a/extras/modules.nix
+++ b/extras/modules.nix
@@ -14,10 +14,10 @@ let
     then module: module
     else
       module:
-      # TODO: set key?
       {
         _class = class;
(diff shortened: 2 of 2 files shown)


roberth:
I'm slightly concerned about overridden inputs not being represented in the key, but this may be the lesser evil of the two.

modules/nixosModules.nix
```diff
@@ -15,7 +15,11 @@ in
       nixosModules = mkOption {
         type = types.lazyAttrsOf types.deferredModule;
         default = { };
-        apply = mapAttrs (k: v: { _file = "${toString moduleLocation}#nixosModules.${k}"; imports = [ v ]; });
+        apply = mapAttrs (k: v: {
+          _file = "${toString moduleLocation}#nixosModules.${k}";
+          key = "${toString moduleLocation}#nixosModules.${k}";
```
roberth, line 20:
```suggestion
          # Note: this neglects to represent potential differences due to input
          #       overrides or flake-parts extendModules. However, the cost for this
          #       is too high or plain infeasible respectively. We choose to implement
          #       deduplication and disabledModules regardless, because not doing
          #       so poses a more direct problem.
          key = "${toString moduleLocation}#nixosModules.${k}";
```

roberth (approved):
Probably best to implement it anyway.

roberth (changes requested):
Coming to think of it, since we have this logic in two places now, it should be factored out, and since we're enabling new behavior, we should test it.

Specifically, `eval-tests.nix` should check that deduplication works (e.g. with a list-typed option that merges by appending, as usual), and that `disabledModules` works.

I won't have time for it in the coming days, so feel free to give it a shot.

liarokapisv:
@roberth Can't we use 
```nix
inputsHash = builtins.hashString "sha256" (                                                                                                    
     builtins.toJSON (builtins.mapAttrs (_: v: v.outPath) self.inputs)                                                                          
);     
 ```
 to also pin the inputs for the key?

PR #262, opened 2024-12-07
Add standard module path compatibility to extras/modules

Pandapip1:


extras/modules.nix
```diff
@@ -66,4 +66,11 @@ in
       apply = mapAttrs (k: mapAttrs (addInfo k));
     };
   };
+  config = {
+    # Copy over to old nixosModules and hmModules attributes
+    flake = {
```
roberth, line 71: I'm a little wary of conditional attributes in the `flake` option, because the condition will have to be evaluated before `self` and `self.outPath` become accessible, producing an infinite recursion otherwise. (Fixing this for `self.outPath` needs a more or less breaking change in flakes / `call-flake.nix`)
In most cases this would be fine, but in the few where it's not, it produces an almost incomprehensible error.

With that in mind, it'd be safer to make this a module that users import separately.

Pandapip1: > With that in mind, it'd be safer to make this a module that users import separately.

Is there any reason that users _wouldn't_ want the standard attribute paths to be populated?

Pandapip1: Instead of a separate module that has to be imported, would an option that disables the behavior an okay compromise?

PR #272, opened 2025-02-22
partitions: allow merging multiple partitions per attr

### LINKED ISSUE #258: partitionedAttrs from multiple sources?
I wonder if it would make sense to support a list for the `partitionedAttrs`. For example if you have multiple partitions providing packages and want to propagate them.

Currently you can do:
```
partitionedAttrs.packages = "dev";
```

But what if you had multiple partitions providing packages?
```
partitionedAttrs.packages = ["component1", "component2"];
```

With this feature you could replace other use cases for subflakes as well.

What do you think, would it make sense to support?

MattSturgeon:
I was looking at refactoring nixvim's flake configuration to take advantage of partitions (https://github.com/nix-community/nixvim/pull/3029), however I noticed that `partitionedAttrs` currently only supports one partition per output attr.

As an example, it might be useful to have the `checks` output attr defined by both `dev` and `docs` partitions, along with maybe some checks defined in the root configuration.

Fixes #258

### DIFF
--- a/extras/partitions.nix
+++ b/extras/partitions.nix
@@ -95,7 +95,7 @@ in
 {
   options = {
     partitionedAttrs = mkOption {
-      type = types.attrsOf types.str;
+      type = types.attrsOf (types.coercedTo types.str lib.singleton (types.nonEmptyListOf types.str));
       default = { };
       description = ''
         A set of flake output attributes that are taken from a partition instead of the default top level flake-parts evaluation.
@@ -147,7 +147,12 @@ in
     _module.args.partitionStack = [ ];
     flake = optionalAttrs (partitionStack == [ ]) (
       mapAttrs
-        (attrName: partition: lib.mkForce (config.partitions.${partition}.module.flake.${attrName}))
+        (
+          attrName: partitions:
+            lib.mkMerge (
+              map (partition: lib.mkForce config.partitions.${partition}.module.flake.${attrName}) partitions
+            )
+        )
         config.partitionedAttrs
     );
   };


terlar:
There was some discussion regarding this in #258

roberth:
The only way to make the merging lazy is to specify explicitly from which partition each attr must be sourced; otherwise, all partitions that contribute to the merge must be consulted.

For something like this to make sense in terms of laziness, we'd need the overrides to be as specific as `packages.<system>.foo = dev`. That way, if `bar` is accessed, only the root partition needs to be evaluated.

Am I correct to assume those are the kinds of attributes you want to specify?
(Where "want" is relative - the fundamental non-laziness of merging isn't fun)

So we'll need to have two things
1. syntax to specify deep overrides easily
    - preferably structured in a way that supports laziness: a tree promotes laziness whereas a list of mappings tends to incentivize conditionals that must be evaluated right at the root. I don't know if these definitions will have conditionals, or how prevalent that is, so grain of salt?
2. syntax to specify overrides in system-specific attrs.
    - Can only consult the root evaluation for the list of systems (again, enumerating systems from partitions would make it strict in all partitions), so may as well implement it in `perSystem` so that it's nicer to use. Is `withSystem` or similar exposed as an option? Will need that, because we only get to access `config` from the partitions. Have an option that contains all module args?

I assume we don't have a real world use case where we want to "wildcard" merge a partition instead of just the root evaluation, because that would be expensive. By making that assumption, we can have a simple syntax like
```
partitions.foo = { };
perSystem = {
  partitionedAttrs.packages.foo = "foo";
  # can't coexist with the above, but that's probably a good thing:
  # partitionedAttrs.packages = "foo";
  
  # I feel like implementing option _also_ in perSystem might lead to a small combinatorial explosion, so let's not, unless proven wrong :)
  # can instead be specified in the non-perSystem option as usual
  # partitions.foo = { };
};
```

I could look into this but I kinda hope it's ok for you to have a simpler setup. I don't want anyone to sink a lot of time into a partitioned setup. Are you sure you have a fetching / locking performance problem?

MattSturgeon:
> The only way to make the merging lazy is to specify explicitly from which partition each attr must be sourced; otherwise, all partitions that contribute to the merge must be consulted.

Ah, thanks for clarifying this. I hadn't really thought this through or tested thoroughly enough before implementing this PR.

Adding some `trace` calls to relevant parts of the flake and its partitions helped me to see in practice how lazy various impls actually are. Not sure if there's a better way to debug laziness & strictness?

> For something like this to make sense in terms of laziness, we'd need the overrides to be as specific as `packages.<system>.foo = dev`. That way, if `bar` is accessed, only the root partition needs to be evaluated.
> 
> Am I correct to assume those are the kinds of attributes you want to specify? (Where "want" is relative - the fundamental non-laziness of merging isn't fun)

Nixvim's flake architecture wasn't really designed with partitioning in mind, so I was trying to fit a square peg into a round hole. In particular, we have a set of docs packages, to which we pass one of our flake inputs. However, that input is only needed for the HTML docs and not the manpage docs, so I wanted to partition the input. I'll postpone refactoring this for future work, for now I'll leave that flake input as a root dependency to keep things simple.

The only other oddity in nixvim is that most of our package outputs relate to the docs, however one is a dev package defined within the `dev` partition. I can work-around that as follows:

```nix
{ lib, config, partitionStack, ... }:
{
  # For any output attrs normally defined by the root flake configuration,
  # any exceptions must be manually propagated from the `dev` partition.
  #
  # NOTE: Attrs should be explicitly propagated at the deepest level.
  # Otherwise the partition won't be lazy, making it pointless.
  # E.g. propagate `packages.${system}.foo` instead of `packages.${system}`
  perSystem =
    { system, ... }:
    {
      packages = lib.optionalAttrs (partitionStack == [ ]) {
        # Propagate `packages` from the `dev` partition:
        inherit (config.partitions.dev.module.flake.packages.${system})
          list-plugins
          ;
      };
    };
}
```

Perhaps this is a niche enough scenario that we don't need to add anything to flake.parts? Otherwise maybe it'd make sense to have a `perSystem.partitionedAttrs` option, rather than implementing some custom syntax?

> I could look into this but I kinda hope it's ok for you to have a simpler setup. I don't want anyone to sink a lot of time into a partitioned setup. Are you sure you have a fetching / locking performance problem?

Agreed. I can solve this for my usage by postponing the docs input being optional/partitioned and explicitly propagating the one package defined by the `dev` partition.

I'll close this PR, as the design is fundamentally flawed. But thanks for your help and insight.

PR #273, opened 2025-02-24
partitions: allow overriding the flake-compat src

MattSturgeon:
This is useful when flake-compat is already available, e.g. as a flake input. By allowing consumers to specify the flake-compat to use, they can avoid fetching multiple redundant copies.

It feels like this could be managed with flake inputs and the `follows` system, however we only depend on flake-compat when a) the partitions module is used, b) a partition has a path-type flake, and c) that partition is actually evaluated. Using a follow-able input would make flake-compat an unconditional dependency. Therefore, I've introduced an internal option that can be used instead (`flakeCompatSrc`).

Maybe the option doesn't need to be internal. I've added the appropriate docs in case it is marked as public (or someone reads the source code). However, the use of `flake-compat` vs `builtins.getFlake` feels like an internal implementation detail, so I was hesitant to make it a public/visible option.

### DIFF
--- a/extras/partitions.nix
+++ b/extras/partitions.nix
@@ -8,6 +8,10 @@ let
     types
     ;
 
+  inherit (lib.generators)
+    toPretty
+    ;
+
   partitionModule = { config, options, name, ... }: {
     options = {
       extraInputsFlake = mkOption {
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Maybe this is conceptually flawed too, since _most_ of the time a flake shouldn't have flake-compat in its unconditional root flake inputs.

So the existing behavior of fetching flake-compat _only_ when evaluating a partition is probably best. This new option may encourage consumers to specify a superfluous flake-compat input in their flakes :thinking:

folliehiyuki:
Another approach is to vendor the relevant part of `flake-compat` into this repo. It's just a single file (less than 300 LOC).

folliehiyuki:
I think this PR can be closed since https://github.com/hercules-ci/flake-parts/pull/315 was merged.

roberth:
Thoughts:
- After https://github.com/NixOS/nix/issues/7730, `flake-compat` should probably just be a flake input like any other.
- I'm open to revisiting this if we have a reason to make it overridable again.

PR #283, opened 2025-05-03
Add quote by Pol Dellaiera to readme

mightyiam:
If it were my project I would have added this quote. It's a matter of taste. Some may want to err on the side of humility.


drupol:
Thank you guys :)

PR #284, opened 2025-05-12
add flake.darwinConfigurations

Mic92:


roberth:
I made a deliberate decision to remove `darwinModules` (ie all nix-darwin-specific attributes at the time), to avoid turning flake-parts into bloatware, to put it bluntly.
Adding to `all-modules.nix` means that each flake evaluates it, and, as a consequence of laziness requirements, it always adds the `flake.*` output attributes, even if they don't have any definitions.

See https://github.com/hercules-ci/flake-parts/blob/c621e8422220273271f52058f618c94e405bb0f5/ChangeLog.md?plain=1#L52-L56

`nixos*` options should get the same treatment at some point. `flake-parts` is not a NixOS framework. We have third party modules that do a better job at that, and decentralization is the whole point of flakes.

As a workaround until it's merged in nix-darwin, you could host this module in clan instead.

PR #285, opened 2025-05-14
Add `#flakeModules.flakeModulesSansAlias`

mightyiam:
Importing `flakeModules.flakeModules`
results in an always-failing `nix flake check`.

This seems to be because it adds a value to
`flake.flakeModules.flakeModule` that ends up as a flake output that
fails to evaluate with the error:

```
Renaming error: option `flakeModules.default' does not exist.
```


roberth:
That would be caused by

https://github.com/hercules-ci/flake-parts/blob/c621e8422220273271f52058f618c94e405bb0f5/extras/flakeModules.nix#L39

To solve this problem, we'd have to either make that behavior optional, or remove it entirely.
By making it optional in an automatic way, we'll introduce an infinite recursion when accessing `self.outPath` (or `self` itself, any attr) from within a `flakeModules` definition, breaking directory-based autodiscovery solutions.
By removing it, we cause trouble for users who import `flakeModule`, as it will disappear without an error message that tells them what to do.

The only solution is see is to add an option like `enableFlakeModuleAlias` so that you can disable that attr in your flake.
Or split the module into a new one that declares `flakeModules` and another that imports the new one and adds the alias. Then you can import the new module without the alias.

mightyiam:
Split module sounds :trophy:  to me.

removed-user:
> That would be caused by
> https://github.com/hercules-ci/flake-parts/blob/c621e8422220273271f52058f618c94e405bb0f5/extras/flakeModules.nix#L39

Are you sure?

Looking through the code it seems to be that the problem is that 
mkTransposedPerSystemModule actually is be passing `flake.flakeModules`

into flake.modules, causing "eager evaluation", on what's supposed to be a deferred module type?


    # top-level flake attribute.
    mkTransposedPerSystemModule = { name, option, file }: {
      _file = file;

      options = {
        flake.${name} = mkOption {
          type = attrsWith {
            elemType = option.type;
            lazy = true;
            placeholder = "system";
          };
          default = { };
          description = ''
            See {option}`perSystem.${name}` for description and examples.
          '';
        };

        perSystem = flake-parts-lib.mkPerSystemOption {
          _file = file;

          options.${name} = option;
        };
      };

      config = {
        transposition.${name} = { };
      };
    };

mightyiam:
@removed-user I can't imagine how this would have anything to do with `perSystem`.

I agree with robert that `.flakeModules` at all should be subsumed in `flake.modules`. So I'm closing this.

removed-user:
> @mightyiam  I can't imagine how this would have anything to do with `perSystem`.
> 
Your're right it doesn't... 

I misinterpreted how it "would" interact with flakeModules, if they were currently under flake.${name}

Because the docs for flakeModules on the flake-parts website show them as something that's defined under flake already; even though they're not.

https://flake.parts/dogfood-a-reusable-module.html
This was where all of my confusion in raising the below issue came from. 
#360

But since the plan is to move them there; 

Then 1: I guess the docs can become correct again?

2: This is relevant anyway, because that would cause issues?

It's not about persystem, it's about the way that the transposition is being done. 
transposing all
flake.${name} items into persystem, that would "possibly" cause evaluation errors when moving flakeModules under flake. 

Because flakeModules are meant to be deferred, and are able to have "raw values", feeding them through persystem, and by-extension... the rest of the module system - could cause premature evaluation of them.

Which would likely "almost definitely" cause infinite recursion.

So, I'm suggesting a filter function before the transposition that can read the flake.name items, and dispatch them to be handled by appropriate transposing functions.

mightyiam:
> transposing all flake.${name} items into persystem

I don't think that that ever happens

> ... feeding them through persystem

I don't think that that ever happens

Issue #288, opened 2025-05-26
`formatter` output is unnecessarily evaluated for all systems


jfly:
## The issue

I noticed that the `formatter` output is eagerly evaluated for all systems my flake supports. For example:

```console
./ $ cat flake.nix
{
  inputs = {
    flake-parts.url = "github:hercules-ci/flake-parts";
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  };

  outputs =
    inputs:
    inputs.flake-parts.lib.mkFlake { inherit inputs; } (
      { ... }:
      let
        currentSystem = "x86_64-linux"; # Set this to the architecture of your dev machine.
      in
      {
        systems = [
          "x86_64-linux"
          "aarch64-darwin"
        ];
        perSystem = { system, pkgs, ... }:
          {
            formatter = assert system == currentSystem; pkgs.hello;
            # packages.default = assert system == currentSystem; pkgs.hello;
          };
      }
    );
}

./ $ nix flake show
warning: Git tree '/tmp/tmp.jM6Snmt11c' is dirty
git+file:///tmp/tmp.jM6Snmt11c
└───formatter
error:
       … while calling the 'head' builtin
         at /nix/store/80jiav8nl8hsx2sf4miaihfacajn1i5h-source/lib/attrsets.nix:1574:11:
         1573|         || pred here (elemAt values 1) (head values) then
         1574|           head values
             |           ^
         1575|         else

       … while evaluating the attribute 'value'
         at /nix/store/80jiav8nl8hsx2sf4miaihfacajn1i5h-source/lib/modules.nix:927:9:
          926|     in warnDeprecation opt //
          927|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          928|         inherit (res.defsFinal') highestPrio;

       … while evaluating the option `flake.formatter':

       … while evaluating definitions from `/nix/store/8lw7ywpxvmmzryn9mnipqhvac4vpwzik-source/modules/formatter.nix':

       … while evaluating the option `perSystem.aarch64-darwin.formatter':

       … while evaluating definitions from `/nix/store/f4g8siq092kgkysvbvaxm5wrclmbmm33-source/flake.nix, via option perSystem':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: string '"aarch64-darwin"' is not equal to string '"x86_64-linux"'
```

<details>
[...]

Issue #114, opened 2023-02-01
Auto-wiring


roberth:
Modules can provide various levels of integration, where an integration is any definition that isn't of a function that it declares.
Sometimes, integrations are not desirable. Making integrations `lib.mkDefault` may help, but this may not always be sufficient. If it turns out that some flakes need to disable too many integrations explicitly, perhaps it makes sense to introduce a single global option that disables all integrations. All (reusable) modules should then adhere to that option.

The story of a flake would then be
1. initialize a flake with flake-parts
2. add some modules with integrations
3. use the integrations
4. add a module with undesirable integration
5. configure that module to disable that integration (?)
6. encounter another undesirable integration
7. set `autoWire = false`
8. write the missing integrations that they did need

Perhaps modules could list their integrations in another option, to help with (8), but all in all it's not a great experience. I would consider this idea to be a last resort; I hope we'll find better solutions.

srid:
Another: https://github.com/Platonic-Systems/process-compose-flake/issues/27

roberth:
Projects (TBD) might provide an alternative
- https://github.com/hercules-ci/flake-parts/issues/205

srid:
Another: https://github.com/juspay/rust-flake/pull/13

srid:
Another: https://nixos-unified.org/autowiring.html

PR #289, opened 2025-05-27
Fix `formatter` laziness and add touchups module

roberth:
This PR is a two-parter by necessity, as the goal is to address
- #288

Unfortunately Nix puts us between a rock and a hard place, and we need more flexibility to deal with possible fallout, hence the new `touchup` feature.

# 1. Strictness problem

To put it briefly, `filterAttrs` induces strictness, but we need it to remove `null`s which `nix flake check` does not allow.
Analyzing `perSystem` is tricky, but doesn't seem to be much of a problem in this particular case, at least in my testing.

- Addresses #288
  - complete resolution involves allowing `null` in `nix flake check`
  
# 2. `touchup` module
  
This might be kind of exciting actually.

It adds a little attribute-based DSL that makes it easy to make alterations to the flake outputs.
A presentation layer of sorts, bridging the normally small gap between Nix-as-programming and Nix-as-data.

Actually it's not attribute-based but (deferred) module based, as illustrated by the last example in `extras/touchup.nix`, which is the file that serves as the introduction (will be published on flake.parts of course).


roberth:
Fortunately this solves the last occurrence of `filterAttrs` that actually processes any kind of input data, indicating it's the last strictness problem of its kind :tada: 

The last remaining `filterAttrs` processes an option whose value belongs to the types / schema / metadata realm.

srid:
While doesn't seem identical, this reminds me of #114 

image

Issue #290, opened 2025-06-01
Using `flake.modules.flake.foo` to use flake module on self


locallycompact:
I assumed this would work from using the flake.modules setting and importing it to self, but it doesn't. Is there a correct way to do this?

https://github.com/cardano-scaling/hydra-coding-standards/pull/5/files

roberth:
> pull/5/files

I expect that to result in an infinite recursion.

I think [Dogfood a Reusable Module](https://flake.parts/dogfood-a-reusable-module.html#dogfood-a-reusable-module) might be of help.

Let me know if you have any more questions.

roberth:
I believe it is possible with this new function, if you could give that a try?
- https://github.com/hercules-ci/flake-parts/pull/291

Site update tbd (separate repo)

The test case is one example, but you could call `flake-parts-lib.importAndPublish` equally well from within a module that's loaded by `import-tree`.
Be careful about which set of module arguments you use, as you can have both the local and caller context available.
```nix
local@{ flake-parts-lib, ... }:
{
  imports = [
    (flake-parts-lib.importAndPublish "dogfood" (caller@{ lib, config, ... }: ... ))
  ];
}
```

By importing a file instead of an anonymous module you can control that a bit.

mightyiam:
Sorry, I don't follow. Perhaps an example?

locallycompact:
The question was how would I do this

| By importing a file instead of an anonymous module you can control that a bit.

mightyiam:
I couldn't understand what "that" means in "control that a bit". So don't know how to answer that. But if the question is how to use that helper with a file, I guess:

```nix
flake-parts-lib.importAndPublish "dogfood" ./path-to-module.nix
```

locallycompact:
If it is a separate file then where will it get the `local` arguments from? And import-tree will also try to import it because it's a file, again having references to `local` arguments which will not exist because there's no binding. I think to do that pattern the `local` arguments need to find their way into the module itself.

locallycompact:
No I think making special exceptions for files undermines the benefit of this approach.

PR #291, opened 2025-06-02
Add `flake-parts-lib.importAndPublish`

roberth:
Context:

- Alleviate #290
- Related: https://github.com/hercules-ci/flake-parts/pull/285#issuecomment-2930158865

### DIFF
--- a/dev/tests/eval-tests.nix
+++ b/dev/tests/eval-tests.nix
@@ -169,6 +169,22 @@ rec {
       partitionedAttrs.devShells = "dev";
     });
 
+  dogfoodProvider = mkFlake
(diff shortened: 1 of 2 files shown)


mightyiam:
If I understand correctly this helper is not strictly necessary; it serves as kind of an instructional material plus a bit of convenience? It other words an alternative to this helper could be some documentation on how to dogfood a flake module ("not by self-reference"). In that case I'm not sure this helper justifies its existence and perhaps education is the better alternative?

locallycompact:
Could we merge this? It is certainly an improvement so thank you.

PR #300, opened 2025-07-31
Add option perSystem.<>.bundlers

### LINKED ISSUE #159: `bundlers` output
That old forgotten api, the [bundlers](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-bundle.html) are missing from `perSystem` 

```
{
  perSystem = {pkgs, ...}: {
    bundlers.identity = drv: drv;
  };
}

  ...
The option `perSystem.x86_64-linux.bundlers' does not exist
```

Lykos153:
Closes #159

### DIFF
--- a/dev/tests/eval-tests.nix
+++ b/dev/tests/eval-tests.nix
@@ -83,6 +83,18 @@ rec {
       };
     };
 
+  bundlersExample = mkFlake
+    { inputs.self = { }; }
+    {
+      imports = [ flake-parts.flakeModules.bundlers ];
+      systems = [ "a" "b" ];
+      perSystem = { system, ... }: {
+        packages.hello = pkg system "hello";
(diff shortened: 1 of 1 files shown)


roberth:
Ok, ignore the test failure. I think what we need to do here is make this an optional module, like `partitions` and whatnot.
That way we avoid adding more "attribute bloat" to all flakes. Let me explain briefly.

Basically we can only pick two of
- easiest way to add a bundler: just add a definition for the option, no `imports` required
- no `bundlers` output attribute when it would be empty anyway
- no surprising infinite recursions when `self` is accessed

I should probably describe this trade-off in more detail in the docs, but anyway, for now I think what matters is that bundlers are fairly rare, so requiring `imports = [ flake-parts.flakeModules.bundlers ];` has the lowest cost of the three.

PR #310, opened 2025-10-18
docs: Clarify internal vs. external import in template

vorburger:
I'm a Nix noob and just learning, but was a little bit confused about this at first; perhaps this helps to clarify it for others who will come along in the future.


roberth (approved):

roberth:
Thanks!

PR #319, opened 2025-11-10
`nixosModules`: class and file

roberth:
Tested #269

### DIFF
--- a/modules/nixosModules.nix
+++ b/modules/nixosModules.nix
@@ -15,10 +15,10 @@ in
       nixosModules = mkOption {
         type = types.lazyAttrsOf types.deferredModule;
         default = { };
-        apply = mapAttrs (k: v: { 
+        apply = mapAttrs (k: v: {
           _class = "nixos";
-          _file = "${toString moduleLocation}#nixosModules.${k}"; 
-          imports = [ v ]; 
+          _file = "${toString moduleLocation}#nixosModules.${k}";
+          imports = [ v ];
         });
         description = ''
           NixOS modules.


cnsta:
This broke my configuration. Nixos modules still work, but not home-manager modules.
```
error: The module `/nix/store/fk957jp0q0nwnvp6idgv81b2z0nipygf-source/flake.nix#nixosModules.home` (class: "nixos") cannot be imported into a module evaluation that expects class "homeManager".
       Help:
       - Ensure that you are importing the correct module.
       - Verify that the module's `_class`, "nixos" matches the expected `class` "homeManager".
       - If you are using a custom class, make sure it is correctly defined and used consistently across your modules.
Command 'nix --extra-experimental-features 'nix-command flakes' build --print-out-paths '.#nixosConfigurations."kima".config.system.build.toplevel' --no-link' returned non-zero exit status 1.
```

I'm sure it might be a skill issue at hand though; and I'm not using flake-parts as intended :)

roberth:
It got a bit stricter in terms of validation. It's supposed to help prevent a different mistake, but assumes that all `nixosModules.<name>` are intended for NixOS.
I'm sorry that we can't really make the error more helpful for this.

To fix this, use the [`homeModules`](https://flake.parts/options/home-manager.html#opt-flake.homeModules) option for home manager modules.

If you don't need merging you don't need to import the home manager flake parts module (see top of linked page).
```nix
flake.homeModules.home = your module;
```

cnsta:
Thanks for the swift reply, and sorry for my delay. I couldn't get homeModules to work with my configuration for some reason I cannot understand. It builds fine without errors, but the homeModules weren't "activated" even though I reference them all over my configuration, like home.derpaderp.enable = true; etc. The fact that it silently accept these, while also ignoring them seems odd to me. I ended up changing from "nixosModules" to "modules" for everything which seemed to solve it.

roberth:
You could import the [`modules` module](https://flake.parts/options/flake-parts-modules.html) and define [`modules.generic.<name>`](https://flake.parts/options/flake-parts-modules.html#opt-flake.modules), which does not declare a class for the module.

PR #331, opened 2025-12-17
lib: add `modulesPath` to `evalFlakeModule`

yunfachi:
With the addition of `modulesPath`, it will be possible to do the same [as in NixOS](https://nixos.wiki/wiki/NixOS_modules#modulesPath):
```nix
flake-parts.lib.mkFlake { inherit inputs; } ({modulesPath, ...}: {
  disabledModules = [
    (modulesPath + "/nixosModules.nix")
  ];
});
```

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -118,6 +118,7 @@ let
           specialArgs = {
             inherit self flake-parts-lib moduleLocation;
             inputs = args.inputs or /* legacy, warned above */ self.inputs;
+            modulesPath = toString ./modules;
           } // specialArgs;
           modules = [ ./all-modules.nix (lib.setDefaultModuleLocation errorLocation module) ];
           class = "flake";


yunfachi:
Ping @roberth

roberth:
I'd like to move away from using file paths for this kind of thing.
Would #350 work for you?

roberth:
In that case I'd prefer to keep it simple and only offer the attributes.
Thanks anyway, glad we have a solution!

yunfachi:
> I'd like to move away from using file paths for this kind of thing.
> Would #350 work for you?

Yes, this actually looks better, thanks. Maybe should avoid using those paths entirely in the whole codebase and rely only on flakeModules instead? I think this would improve consistency and eliminate cases where a new module is added and imported but forgotten in flakeModules.

roberth:
That seems like a good idea. Would you like to look into that?

yunfachi:
> That seems like a good idea. Would you like to look into that?

Yes, I'll open a PR today or tomorrow.

yunfachi:
> That seems like a good idea. Would you like to look into that?

Well, of course using `self(.flakeModules)` in flake-parts imports will cause infinite recursion...

So, what do you think is better:
1. Make `all-modules.nix` a lambda that takes `flakeModules` and returns a module with imports. `flakeModules` (or `allModules` itself) is passed inside the flake `let-in` as arguments to `./lib.nix`.
2. Remove `all-modules.nix`. Pass `flakeModules` into `./lib.nix`, and define the `allModules` list directly there.
3. Remove `all-modules.nix`. Pass `allModules` module into `./lib.nix`.
4. In my opinion, the best option is to define it in `flake.nix` like this: (#351)

```nix
# flake.nix
let
  builtinModules = {
    foo = ./modules/foo;
    # ...
  };
  extraModules = {
    bar = ./extras/bar;
    # ...
  };
  flakeModules = builtinModules // extraModules;

  lib = import ./lib.nix {
    inherit builtinModules extraModules;
    # ...
  };
  # ...
in
# ...
```

```nix
# lib.nix
# ...
modules = builtins.attrValues builtinModules ++ [ (lib.setDefaultModuleLocation errorLocation module) ];
# ...
```

roberth:
Thanks!

Issue #299, opened 2025-07-22
Should flake-parts.modules set the `key` attribute?


jfly:
In https://github.com/NixOS/nixpkgs/issues/215496#issuecomment-2378745043, @roberth said:

> If you use flake-parts, [this module](https://flake.parts/options/flake-parts-modules) sets a file and key for you.

But I think that is incorrect? See this TODO: https://github.com/hercules-ci/flake-parts/blob/644e0fc48951a860279da645ba77fe4a6e814c5e/extras/modules.nix#L17.

jfly:
@roberth, friendly ping here

kalyanoliveira:
> Should flake-parts set the `key` attribute?

I lk think that it should. When you write modules with the
`flake.modules.class.name` option, you can easily run into scenarios where
module deduplication does not work because `key` is not set.

This is a classic diamond-dependency scenario that was discussed
[here](https://github.com/NixOS/nixpkgs/issues/215496) by maintainer @roberth: a
module A depends on modules B and C, and the latter two each depend on some
module D.

What you would usually do in this scenario is make module A import both modules
B and C, and make modules B and C each import module D. Of course, this means
that we are importing module D twice, which is usually fine because it would
have a `key`. However, if it doesn't, you can easily run into issues (e.g.
setting the value of an option twice because it was set by module D).

Of course, the easy fix to this is to forgo any intentions of making your
configuration idiomatic in that way, and resort to importing module D just once
in module A (and never anywhere else, even in modules B and C).

But that sucks.

It especially sucks when we are hinted at the fact that we can do better:

```
# Excerpt of https://github.com/hercules-ci/flake-parts/blob/main/extras/modules.nix
      { ... }:
      {
        # TODO: set key?
        _class = class;
        _file = "${toString moduleLocation}#modules.${escapeNixIdentifier class}.${escapeNixIdentifier moduleName}";
        imports = [ module ];
      };
```

I've written a one-line patch in
"https://github.com/kalyanoliveira/flake-parts/tree/set-default-value-for-key"
for this:

```nix
[horyzon@laptop-nixos:~/flake-parts]$ git diff main
diff --git a/extras/modules.nix b/extras/modules.nix
index 7902a57..e01a8f6 100644
--- a/extras/modules.nix
+++ b/extras/modules.nix
@@ -22,6 +22,7 @@ let
       { ... }:
       {
         # TODO: set key?
+        key = "${class}.${escapeNixIdentifier moduleName}";
[...]

PR #343, opened 2026-03-25
Gives value to `key` attribute of modules composed via `flake.modules`.

kalyanoliveira:
This is my first PR ever! So I don't really know how to document all of this.

The entire discussion and motivation behind this one-line-change PR can be read in #299.

Basically, in scenarios where modules composed via the `flake.modules.class.name` option end up creating diamond dependencies, de-duplication breaks. This is because (for reasons that I don't know) `key` is not set to some appropriate value. 

Please refer to the entire explanation behind the change that this PR introduces [here](https://github.com/hercules-ci/flake-parts/issues/299#issuecomment-3862443945).

Like I said in [my explanation](https://github.com/hercules-ci/flake-parts/issues/299#issuecomment-3862443945), I'm not familiarized with all of flake-parts' particularities and cannot determine if this breaks any existing functionality. Please let me know if that's the case, and I'll attempt my best to fix it.

### DIFF
--- a/extras/modules.nix
+++ b/extras/modules.nix
@@ -22,6 +22,7 @@ let
       { ... }:
       {
         # TODO: set key?
+        key = "${class}.${escapeNixIdentifier moduleName}";
         _class = class;
         _file = "${toString moduleLocation}#modules.${escapeNixIdentifier class}.${escapeNixIdentifier moduleName}";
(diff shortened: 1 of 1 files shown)


kalyanoliveira:
Ah great seems like #251 already exists

PR #349, opened 2026-05-04
fix(inputs'): do not unnecessarily evaluate formatters

### LINKED ISSUE #348: Using `inputs'.foo` can cause fetching of `foo`'s partitions
I found that `inputs'.foo` behaves a bit surprisingly if `foo` uses flake-parts `partitions`. Basically, it causes evaluation of `foo`'s formatter, triggering unnecessary fetching of `foo`'s dev dependencies and evaluation. (I noticed this because a flake I use had evaluation warnings in its formatter.)

I put together a simple repro of the issue here: https://github.com/jfly/flake-parts-inputs-strictness. You can see there that use of `inputs'` triggers a handful of traces and a warning.

IIUC, the issue is that the `perInput` option [has this type](https://github.com/hercules-ci/flake-parts/blob/5250617bffd85403b14dbf43c3870e7f255d2c16/modules/perSystem.nix#L94):

```nix
types.functionTo (types.functionTo (types.lazyAttrsOf types.unspecified))
```

This means any `inputs'.foo` expression will force evaluation of all `perInput` functions (in order to merge their attrset results together). That interacts strangely with the [`perInput` expression in modules/formatter.nix](https://github.com/hercules-ci/flake-parts/blob/5250617bffd85403b14dbf43c3870e7f255d2c16/modules/formatter.nix#L46-L49):

```nix
perInput = system: flake:
  optionalAttrs (flake?formatter.${system}) {
    formatter = flake.formatter.${system};
  };
```

Which forces us to evaluate `flake.formatter.${system}`, which forces us to [evaluate all formatters](https://github.com/hercules-ci/flake-parts/issues/288).

roberth:
This compromises the ability to `inputs'?formatter`, but that seems like a particularly rare use case.
Instead, we treat `inputs'` as another "programmatic" as opposed to "polished" interfaces, similar to the `config` vs flake attributes distinction.
The programmatic interfaces always prioritize laziness whereas the polished interfaces may choose to omit known-missing things, to a reasonable degree.

- Closes #348

### DIFF
--- a/dev/tests/nix-unit.nix
+++ b/dev/tests/nix-unit.nix
@@ -567,6 +567,31 @@ in
     };
   };
 
+  perInput = {
+    "test: inputs' does not evaluate input flake formatter" =
+      let
+        inputFlake = {
+          _type = "flake";
+          outPath = "/fake";
+          formatter = throw "must not evaluate input's formatter";
+          packages.a.hello = pkg "a" "hello";
+        };
+        self = { _type = "flake"; outPath = "/test"; inputs = { inherit self; dep = inputFlake; }; };
+        result = mkFlake
+          { inputs = self.inputs; }
+          {
+            systems = [ "a" ];
+            perSystem = { inputs', ... }: {
+              packages.default = inputs'.dep.packages.hello;
+            };
+          };
+      in
+      {
+        expr = (result.packages.a.default).name;
+        expected = "hello";
+      };
+  };
+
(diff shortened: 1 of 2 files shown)


jfly:
IIUC, this actually closes (closed) https://github.com/hercules-ci/flake-parts/issues/348

jfly:
And thank you for the quick fix! Works great =)

PR #351, opened 2026-05-04
Replace hardcoded paths with `flakeModules`

yunfachi:
Tested with `nix flake check`, and also in a real case using `disabledModules = [ inputs.flake-parts.flakeModules.flake ]` - everything works as expected.


roberth (approved):
Looks great!

Issue #329, opened 2025-12-11
Extend beyond Flakes


roberth:
Something I've been considering for a while.
Flakes serves flake-parts well, but there's untapped potential in what's currently called `perSystem`.

# Concept and Terminology

## Old perSystem

A very useful framing of this option, which is more broad, is:

**module that constructs a possible world for some software to exist in**

or more succinctly, a *configuration* of multiple things.
From the perspective of a thing (e.g. a package), this would be considered the *environment*, containing such things as `build` and `host` platform, and perhaps a small handful of choices that may or may not affect it.

## Old toplevel

This doesn't mean that toplevel is going anywhere. It has earned a solid place as the location to define things that *have already been subjected to the act of* or never were subject of the act of *to configure*, and are instead the result of configuring: a configuration.

Concrete example: if you want your host to have one configuration, put it in toplevel. If you want your host to be cross compiled *from* potentially multiple *build* platforms, put it in whatever we'll call perSystem after this.

## Babel

So we have a bunch of contention on the terms configure/configuration, which are far too similar for the very distinct usages proposed here.
It sure doesn't help that functions are often named for their result, so the confusion goes in all directions.
Brainstorming

Category 1
old toplevel, absolute space, configured, configuration, configurations, concrete, real, determined, definite

Category 2
configurable, configure, function, possible world, world, abstract, relative frame, platonic, undetermined

# Implementation

Circumstance
- Currently largely flake-based ecosystem
- Very few changes could make `perSystem` usable outside the Flakes context
- A few more would make it first-class

Risks
- Things like `inputs'` are somewhat flakes-oriented.
  - Mitigation: make them opt-in, have module that configures flake-parts to behave more portably
  - Solution: achieve parity. `inputs'` is already a mechanism that solves a flaw in flakes, which is that these configurable spaces don't link together naturally except through the flawed `${system}`
- Confusion.
  - Needs good docs
  - bring nix.parts online.

PR #357, opened 2026-05-31
perSystem: use inputs in perSystem instead of self.inputs

Ladas552:
Change behavior of perSystem modules to not assume that self contains inputs.

This was made to allow use of alternative input methods, other than flake inputs, while still utilizing flakes.

In my example it's tack https://github.com/manic-systems/tack

But for the future flake agnostic flake parts (https://github.com/hercules-ci/flake-parts/issues/329) it could be one of the stepping stones for any pinning solution

I did not check if any modules depend on this behavior, but my config rebuid fine

commit switching to tack while successfully rebuilding the config 
https://github.com/Ladas552/Flake-Ocean/commit/6596e800e4d564f3ff252cf4d20f5a6a75856082

### DIFF
--- a/modules/perSystem.nix
+++ b/modules/perSystem.nix
@@ -1,4 +1,4 @@
-{ config, lib, flake-parts-lib, self, ... }:
+{ config, lib, flake-parts-lib, inputs, self, ... }:
 let
   inherit (lib)
     genAttrs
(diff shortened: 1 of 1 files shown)


Ladas552:
I know this isn't the best pr of all time, but damn it takes a while

PR #375, opened 2026-09-02
default moduleLocation: avoid copy to store.

jbgi:
use string concatenation instead of interpolation
to avoid copying outPath to the nix store
(when not using flake, otherwise its already there of course).


roberth (approved):
Should become relevant for flakes too!

